PR #161, opened 2025-04-21
feat: add Remote-User auth (alt names ForwardAuth, Trusted Header SSO)

astappiev:
I have started an attempt to implemented OIDC, as I previously raised in https://github.com/adityachandelgit/BookLore/issues/77.
But was unable to make both, OIDC and simple username/password auths work together.

Therefore, today I decided to abandon the idea and implement MVP solution.

In this PR, I'm suggesting adding a Remote-User auth (requires booklore is running behind a reverse proxy).

This will work with the following setups:
https://doc.traefik.io/traefik/middlewares/http/forwardauth/
https://caddyserver.com/docs/caddyfile/directives/forward_auth
https://www.authelia.com/integration/trusted-header-sso/introduction/

What do you think?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,25 +10,22 @@ COPY ./booklore-ui /angular-app/
 RUN npm run build --configuration=production
 
 # Stage 2: Build the Spring Boot app with Gradle
-FROM gradle:jdk21-alpine AS springboot-build
+FROM gradle:8-jdk21-alpine AS springboot-build
 
 WORKDIR /springboot-app
 
-COPY ./booklore-api/gradlew ./booklore-api/gradle/ /springboot-app/
 COPY ./booklore-api/build.gradle ./booklore-api/settings.gradle /springboot-app/
-COPY ./booklore-api/gradle /springboot-app/gradle
 COPY ./booklore-api/src /springboot-app/src
-COPY ./booklore-api/src/main/resources/application.yaml /springboot-app/src/main/resources/application.yaml
 
 # Inject version into application.yaml using yq
 ARG APP_VERSION
 RUN apk add --no-cache yq && \
     yq eval '.app.version = strenv(APP_VERSION)' -i /springboot-app/src/main/resources/application.yaml
 
-RUN ./gradlew clean build
+RUN gradle clean build
 
 # Stage 3: Final image
-FROM eclipse-temurin:21.0.5_11-jre-alpine
+FROM eclipse-temurin:21-jre-alpine
 
 RUN apk update && apk add nginx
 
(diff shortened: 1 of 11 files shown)


acx10:
@astappiev Thanks! The approach looks good at a glance. I’ll run the code locally and see how it performs. I’ll get back to you soon.

acx10:
@astappiev I had a quick question while going through your approach. Would you mind sharing the specific use case you’re targeting with the Remote-User auth? I’m not very familiar with the inner workings of SSO, so it would be helpful to understand how you expect Booklore to behave in that setup.

Could you also explain how this Remote-User auth differs from simple username/password authentication, and what problem it’s aiming to solve?

Really appreciate the effort you’re putting into this!

astappiev:
I have an average home lab, where I host some apps for my family. And they are exposed to the internet (yeah, playing with a fire).

I can't describe the full setup, but one important part is a reverse proxy.

Which is responsible for:
- routing (it forwards requests to different containers and VMs depending on the request's host)
- filtration (combined with CrowdSec AppSec)
- authentication (I use a default middleware to hide everything behind auth)

Every family member has their account, usually with Passkey (via Authelia). Whenever they want to access any of the hosted apps, they need to use their passkey to authenticate, and after that, they get access to the app. **And after his step, Remote-Auth come into play, it picks up the headers set by reverse-proxy and auto-login to the respective user.**

As I previously mentioned, most, if not all, reverse proxies support this kind of auth:
- https://doc.traefik.io/traefik/middlewares/http/forwardauth/
- https://caddyserver.com/docs/caddyfile/directives/forward_auth
- https://www.authelia.com/integration/trusted-header-sso/introduction/
- https://thehomelab.wiki/books/dns-reverse-proxy/page/setup-authelia-to-work-with-nginx-proxy-manager

In summary, this is required to secure the app with reverse proxy auth middleware and forward the user to automatically login into the app avoiding double authentification and double account management.

The alternative solution, would be to implement OIDC authentication, but as I mentioned, I wasn't able to make it work.

The downside of this method (or advantage), is that any request to the app has to go through the middleware. If we later support some APIs for external clients integrations (e.g. OPDS API, [like in Kavita](https://wiki.kavitareader.com/guides/features/opds/)), they will not be working by default, one would need to add this path prefix (e.g. `/api/opds`) to exceptions in the middleware.

acx10:
Thanks for the detailed explanation! I now have a clearer understanding of your home lab setup and how you’re using the reverse proxy with authentication and auto-login for your apps. The approach you’ve outlined seems solid, and it doesn’t appear to break any existing functionality. It also addresses your use case effectively and could potentially help other users with similar setups. I’m comfortable with this PR and will approve it.

adityachandelgit (approved):

PR #287, opened 2025-05-17
fix: firefox layout issue

afairgiant:
Adjusted css for firefox right hand menu being cut off.

### DIFF
--- a/booklore-ui/src/assets/layout/styles/layout/_responsive.scss
+++ b/booklore-ui/src/assets/layout/styles/layout/_responsive.scss
@@ -35,7 +35,7 @@
 
         &.layout-static {
             .layout-main-container {
-                padding-left: 250px;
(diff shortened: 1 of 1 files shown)


booklore-ui/src/assets/layout/styles/layout/_responsive.scss
```diff
@@ -35,7 +35,8 @@
 
         &.layout-static {
             .layout-main-container {
-                margin-left: 208px;
+                padding-left: 250px;
```
acx10:
```suggestion
                padding-left: 255px;
```

acx10: A 255px offset balances the left margin of the center card, aligning it symmetrically.

adityachandelgit (changes requested):

PR #301, opened 2025-05-18
Fix EPUB Cover Image Detection

afairgiant:
Summary:
This update improves how cover images are detected when uploading or scanning new EPUB files. Some books did not show a cover because the cover was not defined using the expected format.

Changes:

    In EpubProcessor.java, updated the generateCover method:

        First tries epub.getCoverImage() as before.

        If that fails, it searches all resources for any image with "cover" in the ID or file name.

Bug fixed:
Cover images are now shown even for EPUBs that use <meta name="cover" ...> instead of guide or id="cover".

Tested:
Uploaded multiple EPUBs with different cover formats. Confirmed that covers now show correctly, and existing functionality still works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ out/
 local/
 
 booklore-api/src/main/resources/application-local.yaml
+
+docker-compose.yml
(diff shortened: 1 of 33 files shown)


acx10:
Thanks for the fix! That fallback logic makes a lot of sense

PR #309, opened 2025-05-18
fix: restore remote auth

astappiev:
Should fix https://github.com/adityachandelgit/BookLore/issues/303

I'm not sure if there is a better way to work around `router.navigate(['/dashboard']);` in `booklore-ui/src/app/auth-initializer.ts`, but it works now.

This issue https://github.com/adityachandelgit/BookLore/pull/209 persists, so I needed to change some dependencies related to `eslint`.


acx10:
Thanks for the fix! I’ll take a closer look and address the linter issue shortly. Appreciate your work on this!

PR #315, opened 2025-05-18
feat: allow specifying sender address in email provider settings

astappiev:
Solves https://github.com/adityachandelgit/BookLore/issues/286, https://github.com/adityachandelgit/BookLore/issues/300

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/model/dto/EmailProvider.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/model/dto/EmailProvider.java
@@ -16,6 +16,7 @@ public class EmailProvider {
     private Integer port;
     private String username;
     private String password;
+    private String fromAddress;
     private Boolean auth;
     private Boolean startTls;
     private Boolean defaultProvider;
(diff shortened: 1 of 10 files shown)


acx10:
```
<ng-template pTemplate="header">
    <tr>
      <th style="width: 6%;">Default</th>
      <th style="width: 10%;">Name</th>
      <th style="width: 15%;">Host</th>
      <th style="width: 6%;">Port</th>
      <th style="width: 14%;">Username</th>
      <th style="width: 12%;">Password</th>
      <th style="width: 12%;">From Address</th>
      <th style="width: 5%;">Auth</th>
      <th style="width: 5%;">StartTLS</th>
      <th style="width: 7%;">Edit</th>
      <th style="width: 8%;">Delete</th>
    </tr>
  </ng-template>
```

```
<td>
  <input *ngIf="provider.isEditing" type="text" [(ngModel)]="provider.fromAddress" class="p-inputtext w-full"/>
  <span *ngIf="!provider.isEditing">{{ provider.fromAddress }}</span>
</td>

<td class="text-center">
  <p-checkbox
    [(ngModel)]="provider.auth"
    [binary]="true"
    [disabled]="!provider.isEditing">
  </p-checkbox>
</td>
```

I’d prefer to place “From Address” before the Auth and StartTLS checkboxes in the table for better visual alignment.

p2bauer:
Just a suggestion - might be advantageous to bind the from to the destination email to work better with multi-user support? (ie. send-to-kindle for different accounts expect different senders?)

But this is great already, thank you!

astappiev:
@adityachandelgit I changed the order of fields in a form.
@p2bauer I'm not sure if this makes sense. In the current way email providers are implemented, you can add multiple providers with the same credentials, but with different from addresses.

I think the providers/recipients part should be re-evaluated, by first, defining requirements and considering multi-user scenarios in self-hosted app.

Issue #356, opened 2025-05-25
[Bug] Item Count Not Fitting Within Sidebar


royaltongue:
The library "Educational" has 2000+ items within it, but the number is getting cut off by the end of the sidebar. Even so, neither the "Unsorted" library or the "Unshelved" section have this issue. Does it have to do with the length of the library name?

![Image](https://github.com/user-attachments/assets/78fae8f8-59f9-4389-b45d-3a87e811a4ef)

Pweekles:
I am experiencing this issue as well.

BookLore v0.28.0
Hosted on UnRaid docker

![Image](https://github.com/user-attachments/assets/289c6f61-ff32-4196-bdfd-24a6d24bea40)

PR #357, opened 2025-05-25
Allow sidebar count badge to auto-size for larger numbers

afairgiant:
Based on issue #356 
**What’s changed?**

    Dropped the hard-coded 30px badge width so counts can grow.

    Added min-width: 4ch to the count button so 1–4 digit numbers (or more) never get clipped. 

**Why?**
Some library counts were three-digit (or up to four-digit) and were getting cut off in the sidebar. This tweak lets the badge size itself to its content.

**How to test?**

    Inspect the badge in DevTools and swap in a “1234” label to confirm it expands.

    (Optional) Add a dummy library with bookCount$: of(1234) in initMenus() and refresh to see it live.


acx10:
Screenshot 2025-05-26 at 11 36 59 PM

Thanks for the fix! Noticed that the badge content is expanding, but the numbers seem a bit misaligned vertically now. Might need a quick tweak to the padding or line-height. Appreciate all your PRs, really helpful stuff!

afairgiant:
Good catch! I'll look at it again

PR #415, opened 2025-06-01
added series info to book details page

afairgiant:
Added the series and book number to the book details page.


acx10:
Series info is already there

Please check link 27.
`<p *ngIf="metadata.seriesName" class="italic">{{ metadata.seriesName }} #{{ metadata.seriesNumber }}</p>`

afairgiant:
I'm blind apparently 😅

PR #577, opened 2025-06-26
PWA support

### LINKED ISSUE #204: Pwa support
Would it be possible to add pwa support with a manifest? 
The goal is to be able to "install" the app on the homescreen of mobile devices.

klementng:
Add `@angular/pwa` package to allow for installation of an web app on mobile/desktop. 

Resolves #204


domgregori:
With adding PWA support, would be a great feature to have  offline support. Store any books currently reading on device and be able to read it offline. Then when back online, update progress.

I found https://www.flowoss.com which is able to do this, as an example

klementng:
> With adding PWA support, would be a great feature to have offline support. Store any books currently reading on device and be able to read it offline. Then when back online, update progress.
> 
> I found https://www.flowoss.com which is able to do this, as an example

I don't really know what I'm doing, but been trying out a few things here and there to make it work offline. So far, as a proof of concept I've just decided to cache everything via the service worker (it probably should be refined) and made some change to open a book in same tab (rather than a new window, to prevent PWA from opening a new tab in actual browser) and everything seem to be working, with the previously loaded book opening just fine. 

No idea how to update progress when back online/offline (though once you back online and press the next page the progress is likely saved)

klementng:
This pull request for PWA should be somewhat usable offline now. The image for this PR is available here for testing: ghcr.io/klementng/booklore-app:v0.0.0-develop-a88367b (you have to uninstall any prev pwa)

related #652 

The back/refresh ui buttons does not exist on mobile PWA as `minimal-ui` option in the manifest is not supported by mobile browsers. Likely, it will require a separate UI button to return back to homepage for other readers (i only implemented the home button for the epub reader) or you can use the phone gesture for backward nav.

Issue #746, opened 2025-07-20
Dark theme variables not applying due to exact class match selector


TaylanTatli:
Root-level CSS variables are defined using the selector `:root[class="p-dark"]`. However, the `<html>` element often has multiple classes applied (e.g. `class="p-dark fnnsuhax idc0_350"`), especially when extensions are involved.

Because `:root[class="p-dark"]` requires an exact match, the CSS variables don't apply in these cases.

Replacing the selector with `:root[class~="p-dark"]` or `:root.p-dark` would make it more robust and allow the styles to work properly even when additional classes are present.

With `:root[class="p-dark"]`             |  With `:root[class~="p-dark"]` or `:root.p-dark`
:-------------------------:|:-------------------------:
![](https://github.com/user-attachments/assets/16e41130-3b96-458a-be1a-773f25d3013b)  |  ![](https://github.com/user-attachments/assets/11f63742-cd9b-4525-ad78-35e53993cd7f)

adityachandelgit:
Great catch!
Would you be willing to open a PR with the fix? It seems like a simple change, and it would be a big help if you’re up for it!

TaylanTatli:
Sure! Just opened it. Thanks.

PR #747, opened 2025-07-20
Fix dark theme detection when <html> has multiple classes

TaylanTatli:
Fix for #746


acx10:
Thanks! I’ll take a look at it shortly.

adityachandelgit (approved):

PR #778, opened 2025-07-26
Add live demo to README

funkypenguin:
As mentioned at https://github.com/adityachandelgit/BookLore/issues/644, this PR simply adds a link to a live demo to the README :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,15 @@ BookLore is a self-hosted web app for organizing and managing your personal book
 
 ![BookLore Demo](assets/demo.gif)
 
+## 💃 Live Demo
+
+Try it for yourself with the [public live demo](https://demo-booklore.elfhosted.com), sponsored by [ElfHosted](https://elfhosted.com) :heart:, using the following details:
+
+* User: `bookgeek`
+* Pass: `elfhosted`
+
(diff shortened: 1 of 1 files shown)


acx10:
The demo looks great and will be really helpful for users wanting to explore the app quickly.

One concern though: currently the demo allows anyone to upload books, which could lead to unintended copyright issues. I’d prefer not to link to a demo from the README unless uploads are disabled or tightly restricted.

Would you be open to creating a demo user with read-only access (no upload or edit permissions)? That way, users can still explore without the risk of copyrighted content being uploaded.

Thanks!

funkypenguin:
Sure, we can do that. I felt that the hourly reset would mitigate the risk of copyright content being uploaded, since any abuser would just have to re-upload it every hour to take advantage. But if you'd prefer, I can setup a demo read-only user?

acx10:
Thanks, a non-upload user would be great. I assume you already know how to create a user without upload permissions. Appreciate your help!

funkypenguin:
OK, the "bookgeek" user is now a download-only user :)

acx10:
Everything looks good!

If you’d like to populate the library with some free classics, check out the excellent selection of public domain books at:
https://standardebooks.org/ebooks

PR #864, opened 2025-08-10
fixing permissions to be able to run springboot with given UID and GID

mgulyas42:


adityachandelgit:
Hello,

I’m going to revert your PR for now. While it correctly sets permissions for newly uploaded files, it introduces an issue when performing operations on files created in previous versions of the app.

As you know, Booklore’s Java application previously ran as the root user, meaning all existing files were owned by root. With your change, the Java app now runs as a non-root user, which results in permission errors when trying to access or modify those older files.

Here’s an example error I'm seeing:

```
2025-08-10T18:47:32.185Z  INFO 1 --- [booklore-api] [       Thread-9] c.a.b.s.m.parser.AmazonBookParser        : Amazon: Found 0 book ids
java.io.FileNotFoundException: /app/data/thumbs/61/f.jpg (Permission denied)
	at java.base/java.io.RandomAccessFile.open0(Native Method)
	at java.base/java.io.RandomAccessFile.open(Unknown Source)
	at java.base/java.io.RandomAccessFile.<init>(Unknown Source)
	at java.base/java.io.RandomAccessFile.<init>(Unknown Source)
	at java.desktop/javax.imageio.stream.FileImageOutputStream.<init>(Unknown Source)
```

Please let me know if you’d be able to work on a fix that handles both newly uploaded files and files created under the old root-owned structure.

Thanks!

acx10:
Hello @mgulyas42,

I’m going to revert your changes for now. While it correctly sets permissions for newly uploaded files, it introduces an issue when performing operations on files created in previous versions of the app.

As you know, Booklore’s Java application previously ran as the root user, meaning all existing files were owned by root. With your change, the Java app now runs as a non-root user, which results in permission errors when trying to access or modify those older files.

Here’s an example error I'm seeing:

```
2025-08-10T18:47:32.185Z  INFO 1 --- [booklore-api] [       Thread-9] c.a.b.s.m.parser.AmazonBookParser        : Amazon: Found 0 book ids
java.io.FileNotFoundException: /app/data/thumbs/61/f.jpg (Permission denied)
	at java.base/java.io.RandomAccessFile.open0(Native Method)
	at java.base/java.io.RandomAccessFile.open(Unknown Source)
	at java.base/java.io.RandomAccessFile.<init>(Unknown Source)
	at java.base/java.io.RandomAccessFile.<init>(Unknown Source)
	at java.desktop/javax.imageio.stream.FileImageOutputStream.<init>(Unknown Source)
```

Please let me know if you’d be able to work on a fix that handles both newly uploaded files and files created under the old root-owned structure.

Thanks!

acx10:
Most probably we should migrate existing file ownerships at container startup to avoid breaking upgrades.
Something like:

```
#!/bin/sh

# Change ownership of app data only if not already owned by target user
if [ -n "${PUID}" ] && [ -n "${PGID}" ]; then
    echo "Fixing ownership for existing files..."
    chown -R ${PUID}:${PGID} /app/data /app/logs /app/thumbs 2>/dev/null || true
fi

# Start nginx
nginx -g "daemon off;" &

# Start Java as target user
exec su-exec ${PUID:-0}:${PGID:-0} java -jar /app/app.jar
```

This way, existing root-owned files from older versions get fixed automatically, and we still run the app as non-root going forward. What do you think?

In the meantime, I am releasing a hotfix so that users aren’t blocked. We can implement a proper fix like this later.

mgulyas42:
Hello @adityachandelgit!

I’m sorry — it wasn’t intended.

I think it wouldn’t cause issues if you hadn’t set `PUID` and `PGID` before. These environment variables didn’t do anything previously, but since they were in the examples, you’re right — we should handle them.

1. Your idea could work, but I’m afraid it’s too risky to run recursively without restrictions. Maybe we could check specific folders and their ownership before changing anything. If someone mounts the wrong folder and we change its ownership, that would be an unpleasant surprise 😅 That's why I don't like to run docker containers as root
2. In my opinion, the better and more secure solution is to use `user: ${PUID}:${PGID}`, but for that, we might need to remove Nginx.
3. Another option is to create our own environment variables and update the examples, e.g., use `USER_ID` and `GROUP_ID` instead of `PUID` and `PGID`. This way we won’t break existing setups, but those who want can configure it.

**Summary**: The best solution is **2nd**, but it’s too big task for me. I can go with 3rd.

What do you think about?
for 3rd solution: https://github.com/booklore-app/booklore/pull/882/

Issue #808, opened 2025-08-03
[Feature Request]: Series Enhancements


upendedshark:
Hi, love the app so far. I have a few suggestions related to series:

1. Use Series Name when collapsing series: currently when collapsing series, the name of the first book is used. It would be great to see the name of the series.An altogether separate section for series would be even better!
2. Separate series view
3. Add additional series universe field. Some series are a subseries of a wider universe. Might be challenging with metadata providers but if the user can manually add them that's fine. Similar to bookfusion

adityachandelgit:
Hey, really appreciate the suggestions! The series name thing when collapsing has been on my mind too, makes total sense. The universe idea’s cool as well, definitely something we can make user-editable. I’ll try to sneak these in soon.

adityachandelgit:
Just a quick update on (1): if you use the series filter (either from the sidebar or in the URL), the series now stays expanded and shows the series name in the header. Hopefully that makes it a bit nicer to browse through.

For (2), I do want to add a dedicated series view. It’s on my list. Until then, filtering by series is the best way to get that focused view.

And for (3), I’d love to hear more. Do you mean like grouping subseries under a broader “universe” tag or hierarchy? Feel free to share an example or two if you have one in mind.

upendedshark:
Hey thanks for the quick follow up!

Series filter is a good workaround for now, and looking forward to seeing a series section in the future!

For 3 I think there could be 2 options.

The main issue is whether the metadata sources have universes or not. Goodreads does for most books I've seen, but hardcover is a mixed bag and I'm unsure about Google and Amazon. Goodreads shows all the books usually under the publication order, but some series have recommended reading order and/or chronological order. 

I'll use the First Law World as an example. In Goodreads, there are 3 series: [First Law](https://www.goodreads.com/series/43644-the-first-law), [Age of Madness](https://www.goodreads.com/series/211497-the-age-of-madness) and [First Law World (the universe)](https://www.goodreads.com/series/150214-first-law-world). First Law World has all the books. Standalone books in the universe only show up in the First Law World series.

**Series Universe: First Law World (In Reading Order)**
Book 1: The Blade Itself
Book 2: Before They are Hanged
Book 3: The Last Argument of Kings
Book 4: Best Served Cold
Book 5: The Heroes
Book 6: Red Country
Book 7: Sharp Ends
Book 8: A Little Hatred
Book 9: The Trouble with Peace
Book 10: The Wisdom of Crowds
Book 11: The Great Change

[...]

PR #879, opened 2025-08-11
Series Enhancements (1st pass)

rubengarciam:
Related to #808 

**Added:**
- New booklore-api service and query to retrieve the existing series
- New models, routes and components to support Series
- Adds a new element in the sidebar to open all existing series
- `series-browser` and `series-card` components mimic the book to browsing and cards components

**Notes:**
- The `/all-series` many just displays a title. The page doesn't include any filtering, searching or view options, buttons/icons. Hence the height of the container differs from the `/all-books` page. It's pretty basic for now
- I haven't been able to match the existing grid padding for the cards. Probably worth reviewing if the same grid classes can be applied for this view too


acx10:
You can get the series information directly from the books state, no need to make additional API calls to the backend.
The book metadata already includes the series details:

https://github.com/booklore-app/booklore/blob/dd7453fae11a91c1c29467324775e476d15575b8/booklore-ui/src/app/book/service/book.service.ts#L25

You can follow similar pattern as shelves.

Issue #592, opened 2025-06-27
Uploaded files are owned by root, even with PGID and PUID entered.


clockwinder:
Uploaded files are owned by root, even with PGID and PUID entered.

My docker environment includes:
```yaml
    environment:
      - PUID=1001
      - PGID=100
```

But the file (and created folder) added via GUI shows root ownership:

```shell
user@home:/Volume4/appdata/booklore/books$ ls -l
total 12
drwxr-sr-x 2 root users 4096 Jun 27 16:11 'Adrian Tchaikovsky'
drwxr-sr-x 2 root users 4096 Jun 27 16:11 'Aiden Thomas'
drwxr-sr-x 2 root users 4096 Jun 27 16:11 'Kate Heartfield'
user@home:/Volume4/appdata/booklore/books$ cd Adrian\ Tchaikovsky/
user@home:/Volume4/appdata/booklore/books/Adrian Tchaikovsky$ ls -l
total 508
-rw------- 1 root root 518137 Jun 27 16:11 'Tchaikovsky, Adrian - Children of Time.epub'
user@home:/Volume4/appdata/booklore/books/Adrian Tchaikovsky$

```

This breaks my usecase.  Is it the intended behavior?

Is there a way we can import files under ownership of the PGID/PUID?

adityachandelgit:
You should add the `user: "${PUID}:${PGID}"` line in your docker-compose.yml under the booklore service. This tells Docker to run the container with the specified user and group IDs, which should ensure uploaded files have the correct ownership.

```services:
  booklore:
    image: ghcr.io/adityachandelgit/booklore-app:v0.30.2
    container_name: booklore
    user: "${PUID}:${PGID}"
    environment:
      - PUID=${PUID}
      - PGID=${PGID}
      - TZ=Etc/UTC
      - MYSQL_ROOT_PASSWORD=****hidden****
      - SWAGGER_ENABLED=false
    depends_on:
      mariadb:
        condition: service_healthy
    ports:
      - "6060:6060"
    volumes:
      - /home/your-user/NAS/Docker/booklore/data:/app/data
      - /home/your-user/NAS/Media/Books:/books
    restart: unless-stopped
```

clockwinder:
I've not looked at how your dockermake functions, but if `user: "${PUID}:${PGID}"` is implimented in the compose, then the entire container runs as that user, so we shouldn't need the `- PUID=${PUID}` or `- PGID=${PGID}` anymore, right?

[...]

PR #882, opened 2025-08-11
Adding USER_ID and GROUP_ID to specify permission without making breaking change

mgulyas42:
It can fix https://github.com/booklore-app/booklore/issues/592

Based on my previous PR (https://github.com/booklore-app/booklore/pull/864), adding new permission logic would break existing systems that use root:root ownership.

With this modification, anyone who wants to use something other than root:root can configure it with `USER_ID` and `GROUP_ID`.


mgulyas42:
@adityachandelgit any updates? I'm using the forked version because I use non-root file system but its painful to being up to date.

acx10:
@mgulyas42 Thank you for the contribution. I’ve been a bit tied up recently, but I’ll pull the changes soon and verify them locally to ensure there are no regressions with existing user permissions. Your work here is much appreciated, as it improves the experience for users running in non-root environments.

acx10:
@mgulyas42 There are some merge conflicts in your PR. Can you please fix those before I merge in your changes? Thanks.

mgulyas42:
Yes I've just merged develop into the feature branch

kudos:
I ended up here when trying to fix permissions issues. I tried solving it for myself, but after getting the app to lower permissions then Nginx got upset.

Imo, in the containerised world the best solution would be to remove Nginx and have the app serve up requests to static files. Anyone who cares can (and likely already does) run their own reverse proxy in front of it where they can do caching or whatever else they like.

Pocket ID is a good example of an app that started off with an approach like Booklore (except with Caddy), and ended up eliminating the need for it.

mgulyas42:
I agree with you! I mentioned the same in  https://github.com/booklore-app/booklore/pull/864

This current PR is a hotfix for handling non root user. 

Removing nginx would be a breaking change, but worth it. I'd be very happy to contribute, but this is a bit too much for me.

PR #891, opened 2025-08-12
feat: Add docker compose to quickly run a dev env

iamntz:
No need to install java, node and all stuff locally, just run `docker compose -f dev.docker-compose.yml up` and you're good to go.

You can customize exposed ports by adding an `.env` file with `BACKEND_PORT` & `FRONTEND_PORT` value set (defaults are 8080 & 4200).

### DIFF
--- a/booklore-ui/package.json
+++ b/booklore-ui/package.json
@@ -25,8 +25,6 @@
     "@stomp/rx-stomp": "^2.0.1",
     "@stomp/stompjs": "^7.1.1",
     "@tailwindcss/postcss": "^4.1.8",
-    "@tweenjs/tween.js": "^25.0.0",
-    "@types/three": "^0.179.0",
     "angular-oauth2-oidc": "^20.0.0",
     "epubjs": "^0.3.93",
     "jwt-decode": "^4.0.0",
--- a/booklore-ui/package.json
+++ b/booklore-ui/package.json
@@ -25,6 +25,8 @@
     "@stomp/rx-stomp": "^2.0.1",
     "@stomp/stompjs": "^7.1.1",
     "@tailwindcss/postcss": "^4.1.8",
+    "@tweenjs/tween.js": "^25.0.0",
+    "@types/three": "^0.179.0",
     "angular-oauth2-oidc": "^20.0.0",
     "epubjs": "^0.3.93",
     "jwt-decode": "^4.0.0",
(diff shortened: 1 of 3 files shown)


booklore-ui/package.json
```diff
@@ -25,6 +25,8 @@
     "@stomp/rx-stomp": "^2.0.1",
     "@stomp/stompjs": "^7.1.1",
     "@tailwindcss/postcss": "^4.1.8",
+    "@tweenjs/tween.js": "^25.0.0",
+    "@types/three": "^0.179.0",
```
mgulyas42: I’m just contributing here, but why did you add these packages? Also, the package-lock.json file is missing.

iamntz: `package-lock.json` is there though ([link](https://github.com/iamntz/booklore/blob/fd17f40576f201f33cc620497dbb2d63735eee9f/booklore-ui/package-lock.json)).

Those are added because without those the compile throws an error (for those are missing). _probably_ there is legacy 🤷

mgulyas42: I was able to build and start front-end without it. In the first part of the package-lock these dependencies still not there. What is the error if you uninstall them?

iamntz: This is odd. Tried again and worked. Removed from the file.

werwolfby: I had to add this packages as well in my PR (#870), without `npm install` fails when you don't have any `node-modules` yet:

https://github.com/booklore-app/booklore/commit/cbc5300e3089c9217778244a961ccfa192f779a6

```diff
@@ -25,6 +25,7 @@
    "@stomp/rx-stomp": "^2.0.1",
    "@stomp/stompjs": "^7.1.1",
    "@tailwindcss/postcss": "^4.1.8",
+    "@tweenjs/tween.js": "^25.0.0",
    "angular-oauth2-oidc": "^20.0.0",
    "epubjs": "^0.3.93",
    "jwt-decode": "^4.0.0",
@@ -38,6 +39,7 @@
    "showdown": "^2.1.0",
    "tailwindcss-primeui": "^0.6.1",
    "tslib": "^2.8.1",
+    "uuid": "^11.1.0",
    "ws": "^8.18.2",
    "zone.js": "^0.15.1"
```

mgulyas42: As I checked `@tweenjs/tween.js` is a peer dependency of `@iharbeck/ngx-virtual-scroller` so its needed. `@types/three` is not needed

werwolfby: `uuid` is needed as well.

iamntz: Well, since it seems there _is_ a problem, maybe we should have this changed in a different PR?

werwolfby:
@iamntz I did a separate PR for the `npm` issue: #903

PR #903, opened 2025-08-13
Upgrade primeng to 20.0.1

werwolfby:
Running `npm install` and then `npm run dev` or `npm run build` always challenging in the repo.

So I've fix the issues, and made sure that everything compiles now.

The UI check is still in progress.


acx10:
@werwolfby One of the bigger hurdles I ran into was that the p-chips component has been deprecated and removed in PrimeNG 20. Because of that, I held off upgrading to v20. I haven’t found a direct replacement for it yet, did you come across anything suitable as a replacement?

werwolfby:
@adityachandelgit, thank you, I definitely missed `Chips`. Fixed now, was a little more verbose, than I expected 😞.

So according to [migration doc](https://primeng.org/migration/v19#compatible): 

> [!NOTE]
>
> ### Deprecated Components
> Components that are deprecated since their functionality is provided by other components.
>
> `Chips` | Use `AutoComplete` with `multiple` enabled and `typeahead` disabled.

This is exactly what I did:

```diff
              <div class="w-full">
-               <p-chips
+               <p-autoComplete
                  [ngModel]="fetchedMetadata[field.fetchedKey] ?? []"
                  [ngModelOptions]="{ standalone: true }"
-                 [disabled]="true">
-               </p-chips>
+                 [disabled]="true"
+                 [multiple]="true" [typeahead]="false" [dropdown]="false" [forceSelection]="false" styleClass="w-full">
+               </p-autoComplete>
              </div>
```

Now it looks correct, and without any errors in console:

image

PR #986, opened 2025-08-25
Persist sidebar filter state when toggling visibility

adityachandelgit:


Aitanuqui:
The issue persists on mobile.

PR #1009, opened 2025-08-28
Fix GHA: return original names for docker tags

werwolfby:
In #1005 I made an assumption that variables were right (which I specified in warning section).

But assumption was wrong. So I can't test this PR on my personal fork (which this names were intendent to solve).

Just revert names usage to original `booklore` and `booklore-app` values.

P.S. I will check the documentation, probably I could find something like organization name for `booklore-app` and GH registry. And `booklore` could be moved to variables.

@adityachandelgit any thoughts?

Anyway this PR should finally fix the GHA.


acx10:
@werwolfby Let’s go ahead and merge this for now, and we can refine it later once you’ve figured out how to pull the org name from a variable.

PR #1010, opened 2025-08-28
Adding a Helm chart to install in Kubernetes

handyman5:
Hello,

I've created a Helm chart for installing Booklore in a Kubernetes cluster. I tried to make it robust and to use the same configuration as the docker compose setup, to make it easier to deal with.

If you are interested in this but would like it presented in a different form (such as a different directory or a separate repo), I'm happy to make whatever changes you like. I'll keep maintaining it on my end, since I intend to use the service myself.

Thanks for this, it seems like a really great tool!


example-chart/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+  tcpSocket:
+    port: http
+
+# This section is for setting up autoscaling more information can be found here: https://kubernetes.io/docs/concepts/workloads/autoscaling/
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 100
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
werwolfby: Don't think that autoscaling make sense here. For regular API it will work, but for things like Bookdrop, etc. could cause an issues.

Probably better to remove at all to avoid confusion.

example-chart/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+  enabled: true
+  image:
+    tag: 11.8.3
+  auth:
+    database: booklore
+    username: booklore-user
+    existingSecret: booklore-mariadb-password
+
+# This will set the replicaset count more information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
+replicaCount: 1
```
werwolfby: The same as autoscaling, I suggest to hardcode it to 1.

Cause from basic point of view it is not supported by booklore. This is single instance application.

example-chart/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+# Declare variables to be passed into your templates.
+
+mariadb:
+  enabled: true
+  image:
+    tag: 11.8.3
+  auth:
+    database: booklore
+    username: booklore-user
+    existingSecret: booklore-mariadb-password
```
werwolfby: Good to have, but I didn't find usage of it.

example-chart/Chart.yaml
```diff
@@ -0,0 +1,29 @@
...
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.38.2"
+
+dependencies:
+  - name: mariadb
+    version: 22.0.*
+    repository: oci://registry-1.docker.io/bitnamicharts
```
werwolfby, line 29: Bitnami started to support arm64, but previously it wasn't support.

Do you know is it supported in this chart?

example-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+              value: "6060"
+            - name: DATABASE_URL
+              value: jdbc:mariadb://{{ .Release.Name }}-mariadb:{{ .Values.mariadb.primary.service.ports.mysql }}/{{ .Values.mariadb.auth.database }}
+            - name: DATABASE_USERNAME
+              value: {{ .Values.mariadb.auth.username }}
+            - name: DATABASE_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Release.Name }}-mariadb-password
+                  key: mariadb-password
```
werwolfby: Should it be `existingSecret` if specified?

example-chart/templates/pvc.yaml
```diff
@@ -0,0 +1,29 @@
...
+    helm.sh/resource-policy: keep
+  labels:
+    app.kubernetes.io/name: {{ include "booklore.fullname" . }}
+  name: {{ include "booklore.fullname" . }}-books
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ .Values.persistence.booksVolume.size }}
```
werwolfby, line 32: I'm going to use existing NFS share, so I don't need this PVC.

I'm going to specify this in `extraVolumeMounts `. This should be optional.

werwolfby:
Looks like MariaDB operator already have arm64 version, gonna try it someday :)

example-chart/values.yaml
handyman5: Sure, this is all in the default config that `helm create` gives you. I can strip it out.

example-chart/values.yaml
handyman5: The intention is that, if you don't have an existing secret, it gets created for you and persisted across chart upgrades. That way you can safely reconfigure the chart without having to manually reapply the secret, and you don't have to create a secret first to use it (although it works fine if you do). Is there a different behavior you'd prefer?

example-chart/Chart.yaml
handyman5: I have no idea, I'm afraid, but if there's an appropriate image it's possible to override the MariaDB image through `values.yaml`.

example-chart/templates/pvc.yaml
handyman5: I'll fix this tomorrow.

example-chart/values.yaml
werwolfby: Cool! I never knew that there is `helm create` Nice! This explains all this similar charts, I thought that community find some common templates 😄

Anyway, I suggest to remove this one, to avoid any confusion.

example-chart/values.yaml
werwolfby: Yes, this is exact behavior that I'm expecting, I just don't see this `.Values` usage.

example-chart/Chart.yaml
werwolfby: Good point

example-chart/values.yaml
handyman5: It's a little confusing and I'm happy to do this differently. The tricky bit is that you can't templatize the contents of `values.yaml`. I have a `Secret` template that creates essentially `{{ .Release.Name }}-mariadb-password`, and my assumption was that `booklore` was going to be the most likely release name when installing the chart. I get that it's unintuitive and I'm happy to change it, but I did some digging and I couldn't find another helm chart that handled this case in a way I liked better. If you're willing to have the installation instructions include "create a database secrete with `kubectl create secret <whatever>` then we can drop this and the secret template piece, and just have the pod rely on the secret we've asserted exists.

example-chart/templates/deployment.yaml
handyman5: This matches the name of the secret that the chart creates, which we expect to also match `values.yaml` but does not automatically do so. As I said above, I'm happy to change this.

example-chart/values.yaml
werwolfby: I will probably could find some examples...

But essentially if you have this value specified, then no secrets should be created, and this secret should be used for `envFrom`, or just `env` with `secretRef` in the deployment.

Which I don't see here.

But part with creating the secret is completely correct. If this value is not specified, then secret should be created. The only problem that I see in current template is here: 

https://github.com/booklore-app/booklore/pull/1010/files#diff-56177bad038308ee7904c6feeb182cb46a3d9f564af8a65e913fc4a3639fdd8bR1-R2

```mustache
{{- $secretName := printf "%s-mariadb-password" .Release.Name }}
{{- $existing := lookup "v1" "Secret" .Release.Namespace $secretName }}
```

Which expects that secret already exists, and it doesn't use name (`existingSecret`) from values and uses `.Release.Name`.

And this part:

https://github.com/booklore-app/booklore/pull/1010/files#diff-53a1d4395431d6a12e2f5638cdeb4f0be6ea2708ffc417d6830fe21b45ae4ebcR74-R78

```yaml
            - name: DATABASE_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ .Release.Name }}-mariadb-password
                  key: mariadb-password
```

Again doesn't use `existingSecret` values.

You pointed it correctly that it expects default name. My idea here if the value `existingSecret` specified, then it should be used and there is no need to generate template. If this value is not specified then generate a new secret and use it in `Deployment`.

werwolfby: Here is the example from `oauth-proxy`: https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/values.yaml#L29

```yaml
config:
  # Add config annotations
  annotations: {}
  # OAuth client ID
  clientID: "XXXXXXX"
  # OAuth client secret
  clientSecret: "XXXXXXXX"
  # Create a new secret with the following command
  # openssl rand -base64 32 | head -c 32 | base64
  # Use an existing secret for OAuth2 credentials (see secret.yaml for required fields)
  # Example:
  # existingSecret: secret
```

It is commented by default cause it need the secret and it will be created if there is no existing one.

And here is the usage: https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/secret.yaml

```yaml
{{- if and (not .Values.config.existingSecret) (.Values.proxyVarsAsSecrets) }}
apiVersion: v1
kind: Secret
metadata:
.... removed for brevity ....
{{- include "oauth2-proxy.secrets" . | nindent 2 }}
{{- end -}}
```

And a lot of usages in `deployment.yaml`. So I think you've got the idea.

example-chart/templates/deployment.yaml
werwolfby: See my points above.

handyman5:
Ok, I think I addressed all of your comments. Thanks so much for the clear feedback and the good examples to work from!

* I removed the autoscaling and replicaCount bits
* It turns out that the mariadb helm chart made a secret anyway, so I removed mine. Now you'll either get that chart's secret (which the deployment will use) or the existingSecret you specify (which the deployment will also use)
* The persistence volumes can be enabled/disabled and support leveraging existing pvc claims where provided or creating new pvcs if not. There's no native NFS support, but you can mark both as `disabled` and then use `extraVolumes` and `extraVolumeMounts` to generate the appropriate config for your use case.

I _think_ this hits all the notes you gave. If there's anything I missed or anything else you want, do please let me know.

example-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,141 @@
...
+            {{- if .Values.persistence.dataVolume.enabled }}
+            - name: data
+              mountPath: /app/data
+            {{- end }}
+            {{- if .Values.persistence.booksVolume.enabled }}
+            - name: books
+              mountPath: /books
+            {{- end }}
+            {{- with .Values.extraVolumeMounts }}
+            {{- toYaml . | nindent 2 }}
```
werwolfby: `nindent 2` - sounds like not enough here, should be `12` I assume (or `10`?).

example-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,141 @@
...
+        - name: books
+          persistentVolumeClaim:
+            claimName: {{ if .Values.persistence.booksVolume.existingClaim }}
+              {{ .Values.persistence.booksVolume.existingClaim }}
+            {{ else }}
+              {{ include "booklore.fullname" . }}-books
+            {{ end }}
+        {{- end }}
+        {{- with .Values.extraVolumes }}
+        {{- toYaml . | nindent 2 }}
```
werwolfby: `nindent 2` - sounds like not enough, should be `8` I assume.

werwolfby (commented):
Thanks a lot for help and time for implement the chart.

It looks good to me, except this 2 questions for `nindent`, I'll give it a try soon.

handyman5:
Good catches on the indents, I always screw up spacing in helm charts. I realized I had too many newlines in the `claimName`s too, so I fixed that as well. Thanks again for all the great feedback.

s3rverro0m:
This looks great! I’ll try the chart out this weekend.

acx10:
@werwolfby I haven’t had a chance to fully review the PR. Have you had an opportunity to make a final decision? Are we good to merge it?

werwolfby:
@adityachandelgit I was trying to install chart myself, but cause it doesn't published and I'm want to have everything in gitops it didn't workout.

So I can't confirm that it works or not. But from code point of view eveything that I see sounds reasonable.

And my advice is to merge it, and let community try it. And fix it on the way if we will find any issues.

Issue #327, opened 2025-05-20
Support different databases


joszz:
I like the looks of this project and I'm interested to set it up.
However I'd like to be able to run it on Postgres instead of MariaDB.
Since your using JDBC it seems pretty trivial to implement other databases (sqlite, mysql, postgres would come to mind).

I tried, and it seems the dependencies for these are missing.

MaheshBabu11:
The challenge here is the query dialect between MySQL and PostgreSQL for example are different. Since flyway is used for db migrations we can sort this by creating  separate scripts for each db which i am not sure is a right approach.

This way for example.

src/main/resources/db/migration/
│
├── V1__Create_Tables__postgresql.sql
└── V1__Create_Tables__mysql.sql

Flyway auto-detects which to use based on spring.datasource.url. 

Other than that its just a matter of adding the gradle dependencies and passing the env variables.

MaheshBabu11:
Hi @adityachandelgit , is there any particular reason why flyway is used ?

Spring jpa hibernate property can be used to create the tables right? 

Skip over the whole db conundrum and just BYO DB ?

adityachandelgit:
Hi @MaheshBabu11, thanks for your input! I don’t plan to support multiple databases because maintaining compatibility across different SQL dialects adds significant complexity. Each database (Postgres, MySQL, MariaDB, etc.) has subtle differences in SQL syntax, functions, and behavior that require separate migration scripts and thorough testing.

Regarding Flyway, it’s used to reliably manage and version schema migrations. While Spring JPA/Hibernate can auto-generate tables, it doesn’t offer precise control or versioning for evolving schemas. Flyway ensures migrations are applied safely and consistently over time.

Since I want to keep the project lightweight and maintainable, focusing on a single database (MariaDB) and using Flyway for migrations keeps things simpler and more reliable. Supporting multiple databases would increase maintenance overhead and slow development.

Thanks for understanding!

fdw:
Hey, I'd like to chime in, because I think that Postgres is more common among self hosters than MariaDB.

Does BookLore use any special features from MariaDB that are not available in standard SQL? If it's possible to use only standard SQL for migrations and queries, it should work with many DB engines; also, JDBC and Spring data already support more or less any you can imagine. That would be a great feature in my eyes 🙂

pr0ton11:
As soon as PostgreSQL would be supported, this could run on my infrastructure. I am not willing to setup MariaDB just for this.

[...]

PR #1095, opened 2025-09-06
feat: Add support for PostgreSQL database

fardyceps:
Hello,

This MR is an attempt at addressing issue https://github.com/booklore-app/booklore/issues/327 by adding support for PostgreSQL database. Specifically this MR adds PostgreSQL compatible migration scripts and moves all MariaDB specific scripts and common into separate folders. Flyway should then be able to automatically pick up which scripts to use for which engine.

Admittedly this approach is a bit cumbersome due to having to maintain different scripts for each database, but it will not break existing installations.


fdw:
I like this idea. Have you tried whether it breaks exisiting installations because the script file name changes?

Also, future migrations should be added in the common dialect, so there's no additional pain.

fardyceps:
> Have you tried whether it breaks exisiting installations because the script file name changes?

Yea I did. Tested against my installation. AFAIK Flyway only validates against the content of the script, not the file name or file path, so renaming/moving them shouldn't be a problem.

pr0ton11:
Commenting to get updated whenever this PR is merged. Thank you very much.

Kernald:
To hopefully avoid more noise - there's no need to comment to get notifications. That's what the "Subscribe" button is for. Similarly, expressing your interest is what thumbs up are for.

ckozak:
wouldn't sqlite be a more fitting database for this service? That would allow the service to be self-contained, and not require an external database.

fardyceps:
> wouldn't sqlite be a more fitting database for this service? That would allow the service to be self-contained, and not require an external database.

Yeah, I understand that SQLite is probably enough for most people. I also prefer SQLite in my services. However my current set up already has Postgres and a lot of people seem to want it in #327 so that's what I chose to write PR for.

People are welcomed to write PR for SQLite. The process would be very similar to this one. I played around with it but I think @adityachandelgit may have already decided that multi database support is not gonna be in the roadmap.

DanielRamosAcosta:
Hello booklore community! @acx10 could you please clarify why this PR was closed? I believe implementing Booklore with standard SQL would enable users to choose their preferred database rather than being locked into MariaDB.

Are you open to accepting new PRs that add multi-database support? I'd be happy to contribute to this effort if it aligns with the project's direction.

acx10:
@DanielRamosAcosta 

Thanks for the interest and for offering to contribute.

Supporting multiple databases (for example, MariaDB and PostgreSQL) isn’t just about using “standard SQL.” In reality, there are real differences in SQL dialects, indexing strategies, JSON support, full-text search, pagination, and performance characteristics. Even when using JPA/Hibernate, these differences inevitably leak into the application code.

Maintaining true multi-database support would significantly increase complexity, testing scope, and long-term maintenance effort. Given the current scope of Booklore and the time I can realistically dedicate to it, I’ve chosen to focus on a single, well-tested database rather than dilute effort across multiple backends.

This isn’t a hard “never,” but multi-DB support would require someone to fully own the implementation, testing, and ongoing maintenance. If that level of commitment exists, I’m open to revisiting the idea.

DanielRamosAcosta:
Thanks for the detailed explanation @acx10, that makes total sense 😀

PR #1123, opened 2025-09-10
bug: Consider series name when sorting books

rubengarciam:
The books sorting now takes into consideration the series name when books are collapsed in series. Beforehand it was still using the firs book series title, displaying the series out of order.

image


rubengarciam:
@adityachandelgit please include this PR before the next release with the new series improvements #1086

PR #1129, opened 2025-09-11
feat: OPDS v2 support for libraries, shelves

rubengarciam:
Adds OPDS to navigate through:
-  libraries
-  shelves
- recently added books (matching dashboard's view)
- as well as the complete catalog. 
 
This does not include _magic shelves_ support yet.

**Important:** The OPDS endpoint to connect to is `http://localhost:8080/api/v2/opds` not `http://localhost:8080/api/v2/opds/catalog`. Pointing to `'/catalog` is the route to return all the existing books in the library catalog

PR notes:
- Compliant with OPDS `application/opds+json` media type requirements
- Includes paginated results support
- Adds the new mappings and logic for the OPDS controller/service classes
- Validates the data returned is only what the authenticated user can access
- Added the required Book Service methods and DB queries
- +Test classes

Commits:
- [2cd8bb2c](https://github.com/booklore-app/booklore/commit/2cd8bb2c7ee0f6a0274e2973b61a2d3e2bdcc86b): improvement foundations + libraries, shelves navigation
- [0aa9a41](https://github.com/booklore-app/booklore/commit/0aa9a41059bea5c0f92567e2536f4dfcd34b2983): recently added books navigation

[OPDS v2 specs reference](https://drafts.opds.io/opds-2.0.html#1-overview)

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/controller/OpdsController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/controller/OpdsController.java
@@ -65,6 +65,14 @@ public class OpdsController {
                 .body(feed);
     }
 
+    @GetMapping(value = "/recent", produces = {"application/opds+json"})
+    public ResponseEntity<String> recent(HttpServletRequest request) {
+        String feed = opdsService.generateRecentFeed(request);
+        return ResponseEntity.ok()
+                .contentType(MediaType.parseMediaType("application/opds+json;profile=acquisition"))
+                .body(feed);
+    }
+
     @GetMapping(value = "/search.opds", produces = "application/opensearchdescription+xml")
     public ResponseEntity<String> searchDescription(HttpServletRequest request) {
         String feed = opdsService.generateSearchDescription(request);
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/BookQueryService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/BookQueryService.java
@@ -52,6 +52,21 @@ public class BookQueryService {
         return new PageImpl<>(mapped, pageable, books.getTotalElements());
     }
 
+    public Page<Book> getRecentBooksPage(boolean includeDescription, int page, int size) {
+        Pageable pageable = PageRequest.of(Math.max(page - 1, 0), size, Sort.by("addedOn").descending());
+        Page<BookEntity> books = bookRepository.findAllWithMetadata(pageable);
+        List<Book> mapped = books.getContent().stream()
+                .map(book -> {
+                    Book dto = bookMapperV2.toDTO(book);
+                    if (!includeDescription && dto.getMetadata() != null) {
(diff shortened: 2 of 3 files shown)


ForceConstant:
I was excited to try this feature out, but it doesn't work for me. I tried just removing the catalog from the url, but then moon reader just gets an error, but it does work with default url with catalog. 

```
At line 1, column 0: not well-formed (invalid token)
{"metadata":{"title":"Booklore"},"links":[{"href"
```

rubengarciam:
@ForceConstant  I am not sure MoonReader supports OPDS v2 (only up to v1.2?)

redikhet:
In KoReader on my Boox the new opds feature does not seem to work correctly.
It seems that syncyng in the background happens, but when clicking on the catalog nothing shows.
/catalog is still working with the old behaviour, but /libraries, and /shelves do not show anything after clicking.

User has access to all libraries so it does not seem a permission issue.
BTW /catalog was not honoring user library permision anyway, but thats another thing.

farwestnz:
My KOReader errors if I remove “catalog” too.

Aitanuqui:
Not working when removing "/catalog".

rubengarciam:
@farwestnz @redikhet I have just found that KOReader does not support OPDS v2. Just up to v1.2.

I'm keeping tabs of the readers that support OPDS v2 here: https://github.com/booklore-app/booklore/issues/1157

redikhet:
> @farwestnz @redikhet I have just found that KOReader does not support OPDS v2. Just up to v1.2.
> 
> I'm keeping tabs of the readers that support OPDS v2 here: #1157

Thanx, it doesn't seem that KoReader has plans to upgrade to V2 soon ☹️

rubengarciam:
@redikhet - adding to my previous comment. The current v2 implementation falls back to v1 if the app doesn't include the `Accept: application/opds+json` header, that's why you receive all the books in the catalog and permissions are not respected. We should probably remove it and just return "not authorized".

+cc @adityachandelgit

redikhet:
> @redikhet - adding to my previous comment. The current v2 implementation falls back to v1 if the app doesn't include the `Accept: application/opds+json` header, that's why you receive all the books in the catalog and permissions are not respected. We should probably remove it and just return "not authorized".
> 
> +cc @adityachandelgit

I hoped that i could select the library by using another user. As a workaround for not having v2, to differentiate the opds streams and only get the data from a specific user that is connected to a library.

farwestnz:
It must be possible to get some improvement without moving to OPDS v2. If I access the OPDS catalog for Kavita from KOReader, I get a top level menu with:

- On Deck (eg Currently reading)
- Recently Updated
- Recently Added
- Unread books
- Reading Lists
- Want to read
- All Libraries
- All Collections
- Smart Filters (eg Magic Shelves)

ForceConstant:
Yes @rubengarciam can we get these library features added for v1?

rubengarciam:
I'll look into it

PR #1166, opened 2025-09-16
update docker-compose to avoid the docker run with blank page, which due to can't connect to db

RocGod:


iamntz:
Ports shouldn't be exposed if not needed, and the containers running in the same compose „namespace” makes port exposing useless. I think the same is true for the network as well.

Most likely the reason you had a white screen is because db migration wasn't completed.

RocGod:
> Ports shouldn't be exposed if not needed, and the containers running in the same compose „namespace” makes port exposing useless. I think the same is true for the network as well.
> 
> Most likely the reason you had a white screen is because db migration wasn't completed.

The issue I am facing is from zimaOS and CasaOS, which port is not found, as we are using jdbc db connection with port 3306, without setup the port in db, how does the container know which port is being used. Also the error log mentioned the booklore can't connect the db. DB container running well. 
This is the reason why I put port and network, even network might not be mandatory.

SagaciousZed:
I don't think its a good idea to expose the DB port by default. I did not have to expose the database port in my setup on a Synology. 

Docker Compose should have already placed the services on the same default network. Containers on the same network do not need ports declared to communicate on those ports.

If you needed the ports exposed on ZimaOS or CasaOS it would seem to be a quirk with them. The booklore container knows where the database is because of this environment variable https://github.com/booklore-app/booklore/pull/1166/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L109

RocGod:
> I don't think its a good idea to expose the DB port by default. I did not have to expose the database port in my setup on a Synology.
> 
> Docker Compose should have already placed the services on the same default network. Containers on the same network do not need ports declared to communicate on those ports.
> 
> If you needed the posrts exposed on ZimaOS or CasaOS it would seem to be a quirk with them. The booklore container knows where the database is because of this environment variable https://github.com/booklore-app/booklore/pull/1166/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L109

Could you please explain what the harm or risk to expose the DB port in a local docker container? As I am facing this issue in ZimaOS, and after I expose this DB port in docker compose, the issue is gone. You already put 3306 in your booklore docker config, any concerns about that?

iamntz:
> Could you please explain what the harm or risk 

As soon as you expose 3306 on a particular container, that port is locked. And sooner or later you will have conflicts with other containers.

Further more, any app on your host (I'm not sure, but I think that all other containers as well) could access your DB by `localhost:3306`

Now, if this doesn't work on ZimaOS, I'd say it's a better approach to investigate why is that, because _most likely_ you (will) have the same issue with other compose files.

SagaciousZed:
While there are currently no specific security concerns with MariaDB. The risk is the increase the increased surface area. This does not adhere to the principle of least privilege, because it allows all external machines that can route to the host to connect to the database container, which is not needed for the intended purpose.

The database address being in the config, is not a concern because the container name should not be routable outside the docker network. To be clear, when the database port is listed under ports, it is listening on a public interface, and is no longer a localhost only container.

RocGod:
> While there are currently no specific security concerns with MariaDB. The risk is the increase the increased surface area. This does not adhere to the principle of least privilege, because it allows all external machines that can route to the host to connect to the database container, which is not needed for the intended purpose.
> 
> The database address being in the config, is not a concern because the container name should not be routable outside the docker network. To be clear, when the database port is listed under ports, it is listening on a public interface, and is no longer a localhost only container.

Thanks for explanation. I just had this issue with zimaOS and figure out this solution. That's the reason I raise this PR. If you think this will cause other problems, or have more risk, you can decline this PR. But still I would like to keep this solution for casaOS/zimaOS user only. Thank you again.

PR #1167, opened 2025-09-16
Filter "Continue Reading" to only show READING, RE_READING, or PAUSED…

adityachandelgit:
… books


maddler:
Think 1.10.0 introduced a regression. This is happening again.
capture-2025-10-28_09-08-04_p m
capture-2025-10-28_09-06-17_p m

PR #1168, opened 2025-09-16
Feat: Add number of books read in Series Page

rubengarciam:
Adds the number of books read from the total in the series page.

Added "READING" status when reading or re-reading at least one book from the series

image


farwestnz:
This is partially working for me. Sometimes the series says "Reading" status with no book count, sometimes it says "Partially read" with the book count. I don't seem to be able to manually set the status at the series level to experiment further...

rubengarciam:
@farwestnz that is by design. The page currently displays `READING` if the user is actively reading any book in the series, and `PARTIALLY READ` if some books are mark as read and others not, without any book been in a reading status.

Applying metadata to all the books in a series in bulk is not yet implemented.

farwestnz:
Thanks for the clarification, it makes sense. Is there a reason we haven't added the book count for `READING` status?

rubengarciam:
@farwestnz no reason in particular, there's definitely a case to add it and defining what is the wording that makes more sense. I've created an issue for this, let's continue the conversation there #1196

Issue #978, opened 2025-08-24
Two users cannot put the same book on their Kobo shelves


SpeakingOfBrad:
# Bug Report Template for Booklore

**What happened?**  
My wife added a few books to her Kobo shelf. I tried adding one of those same books to my Kobo shelf, and as soon as I click save it logs me out and shows an error popup "Error: Failed to update book shelves".

**How can we see it happen?**  
Steps to reproduce the issue:  
Preqrequisites: Two users with Kobo sync enabled.
1. Log in as UserA.
2. Add a book to your Kobo shelf.
3. Log out (or use a new private window) and log in as UserB.
4. Add the same book to your Kobo shelf that UserA did.
5. See error message.

**What did you expect to happen?**  
Two users should be able to put the same book on their respective Kobo shelves. (Otherwise book clubs are going to be difficult.)

**Screenshots or Error Messages**  
<img width="716" height="511" alt="Image" src="https://github.com/user-attachments/assets/325ea421-9f7f-4570-9755-2eb32e781b12" />

**About your setup:**  
- Booklore version
  - v0.38.0
- What computer or device are you using?
  - Ubuntu
- Which browser and version?
  - Firefox  
- How did you install Booklore?  
  - Docker, on Ubuntu VM. Booklore is accessed via subdomain that is reverse proxied with Caddy.

**Anything else we should know?**  
I don't see any errors in the Docker logs, otherwise I'd add those here.

adityachandelgit:
That’s strange, I was able to add the same book to the Kobo shelves of two different users without any issues. Do you see any server logs?

<img width="1733" height="998" alt="Image" src="https://github.com/user-attachments/assets/e55e8713-5e4e-48bd-ad26-3c50d0f867e6" />
<img width="1733" height="998" alt="Image" src="https://github.com/user-attachments/assets/2756f5df-4993-417c-aa89-c63031d7fccb" />

SpeakingOfBrad:
I don't see any new logs when it logs me out. I'm looking in the docker logs for this (`docker logs -f booklore` or `docker compose logs -f`), is there somewhere else I might find logs?  

I'm not sure if this is expected or if it may be a clue, but my user profile has no shelves except for the Kobo shelf and it shows that the "Unshelved" shelf has 4 less than the total. Though my wife does have 4 books on her Kobo shelf.

[...]

PR #1201, opened 2025-09-23
fix: stop sending shelves not associated with the user

jduar:
Fix for #978 

It might make sense to filter the shelves in the book mapper itself so it happens on every use of `toBook`. But given that it's used in a lot of places and I'm not familiar with most of the project's code I'm afraid that could introduce some issue I'm unaware of. So I filtered the resulting shelves specifically in the method needed to fix the existing issue.

I tested it locally with the same setup of users and books with which I could reproduce the error and this fixes it.

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mapper/ShelfMapper.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mapper/ShelfMapper.java
@@ -3,9 +3,11 @@ package com.adityachandel.booklore.mapper;
 import com.adityachandel.booklore.model.dto.Shelf;
 import com.adityachandel.booklore.model.entity.ShelfEntity;
 import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
 
 @Mapper(componentModel = "spring")
 public interface ShelfMapper {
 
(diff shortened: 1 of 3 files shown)


jduar:
After more tests, I think this needs further tweaking. When assigning the shelf with my test user, the error no longer happens but the shelf isn't actually saved. Though the return indicates a 200 and shows the shelf on return, when reloading, it appears as if the shelf hasn't been assigned again. I suspect some place else where this query happens is getting in the way of actually assigning the shelf. I'm moving this into draft until I figure it out.

EDIT: I'll leave this just in case. Going straight to the `book_shelf_mapping` db table I can see that the test book is assigned to all shelves. But it's not showing to the test user for some reason. Something about the previous fix might have broken the user shelf mapping but I can't yet tell what. I'll try to understand this tomorrow.

jduar:
Everything is working now. Shelves weren't being correctly mapped to books and there was another method where there was a need to filter the shelves by the authenticated user.

It might make sense in the future to evaluate which other methods could benefit from this filtering and if it might make sense to do it directly in one centralized place, like the mapper for example. I don't have the knowledge of the project to do this however so I applied the fix to the two methods in the service which I am able to test thoroughly.

Issue #1157, opened 2025-09-15
OPDS v2 Readers


rubengarciam:
Tracking readers and OPDS v2 support

rubengarciam:
Updated with Cantook, FBReader, Moon+, Stump

Iskari:
Cantook seems to support authentication (at least it works on the demo Komga instance, which uses OPDS v2 with Authentication). 
I think Cantook expects the server to implement the OPDS Authentication Spec.

Spec: https://drafts.opds.io/authentication-for-opds-1.0.html
Komga Release Notes with specific commits: https://github.com/gotson/komga/discussions/1476

rubengarciam:
The current OPDS implementation supports basic auth and Cantook should work ok with that. The specific Komga commit seems to add an auth map though, so I'll review the v2 spec (the one added was for v1) and see if there's anything missing.

ForceConstant:
@rubengarciam So I tested on MacOS, and yes Pocketbook is working for v2, but the auth is not working on android. I looked at the raw packets, and it seems that the android app is not actually sending the auth, even though it asks for username/password. i sent an email to let them know.

Kernald:
The current UI is very confusing, but BookLore doesn't support, as far as I can tell, OPDS v2 yet. The current two tabs are two implementations of OPDS v1, with the first version being deprecated. The sooner it's gone the better, because yeah the current state of things, with two versions of OPDS v1, did lead myself to the same concerns before being told they were both OPDS v1...

arcoast:
@Kernald That is very interesting, and now I look at the api endpoints, they are both indeed XML, whereas I believe v2 of the OPDS spec is json based. 

Might pick Booklore up again now, and see if I can get Koreader working with the "v2" API (which is actually v1) 😄

beudbeud:
I love [Mihon reader](https://mihon.app/) for manga cartoon.  There are already a plugin for Kavita and Komga but it will be very nice to have one for Booklore

PR #1236, opened 2025-09-28
Drop legacy OPDS and keep only the new implementation

acx10:


ForceConstant:
I was surprised when I saw this om the changelog as it seemed from the feedback doesn't seem like there were enough clients supported v2 to warrant getting rid of v1

https://github.com/booklore-app/booklore/issues/1157

PR #1254, opened 2025-10-03
Resolve Kobo Library Synchronization Issue

adityachandelgit:


systemallica:
Has this been always broken? Or is this a new requirement? If the latter, I'm curious as to why 🤔

PR #1275, opened 2025-10-06
Feat: set reading status using Koreader Sync Plugin

WorldTeacher:
This small PR is my attempt at setting the read status for a book based on the KoReader Sync plugin. While I am fairly new to Java dev, and therefore currently not set up to test this locally, I assume this should work, but changes are welcome. 

This was created in reference to [this discord feature request](https://discord.com/channels/1379932933581963444/1379934554957152336/1424024474898141255)

I'm not sure if this can be applied to the other types as well, I only have epubs in my library and therefore only added the part to  the epub area

Feedback / Suggestions are welcome


acx10:
Looks great, thanks for the contribution! 

I think KoReader sends progress for both EPUBs and PDFs too, though I’m not 100% sure, but your code should handle both.

PR #1343, opened 2025-10-14
feat: IPv6 compatibility

mhymny:
Makes nginx also listen to IPv6

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -15,6 +15,7 @@ http {
 
     server {
         listen ${BOOKLORE_PORT};
+        listen [::]:${BOOKLORE_PORT};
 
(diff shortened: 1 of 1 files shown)


mhymny:
I have booklore running on my IPv6 only cluster using `BOOKLORE_PORT=[::]:6060`, as that is envsubstituted into the nginx.conf.
It seems to be running well so far.

Aside from Booklore additionally listening to IPv6:
As of now I don't know whether I can recommend single-stack IPv6, as some metadata provider(I think it was amazon.com) are not resolved properly or some thing like that...Additionally github is still IPv4 only so a lookup for the latest release fails.
Will take a look at that.

PR #1423, opened 2025-10-23
Feature/docker improvement

prax93:
- Created .env File for easier Env Variable Handling
- Modified docker-compose file to use .env file
- Updated Readme File


acx10:
@prax93  I’ve already made similar updates in the develop branch, could you take a look? I believe this PR might no longer be needed. Thanks again for contributing!

PR #1438, opened 2025-10-25
Show a generated cover for books which are missing a cover image

cjd:
Currently any book with a missing cover is given a static missing-cover.jpg when displayed.
This change replaces that with a generated/styled SVG which contains the book title and author.
It has minimal styling (just enough to not be boring/stand out too much in a list) and works by adding a cover-generator component.

Here is an example output:

image


acx10:
This looks great, I can’t believe I didn’t think of this earlier! Thanks for putting this together.

Issue #1141, opened 2025-09-13
[Bug] Uploading books is not possible without 'Manage libraries' enabled


SirJohnDoe:
# Bug Report Template for Booklore

**What happened?**  
I've created a new user to test permissions. I noticed that if 'Manage library' is disabled, the Bookdrop and Upload buttons at the top disappear.

**How can we see it happen?**  
Steps to reproduce the issue:  
1. Create a new user
2. Give the user the following permissions:
<img width="719" height="210" alt="Image" src="https://github.com/user-attachments/assets/dfd80bc3-7011-44a2-b467-608575abdb50" />

3. When logging in on the new user, notice that Bookdrop and Upload Book is not visible at the top
4. Add the 'manage library' permission to the user, and Upload and Bookdrop show up at the top.

**What did you expect to happen?**  
The 'Upload' permission by itself should enable the Upload Book button at the top. The manage library permissions gives a lot of unwanted access.

**About your setup:**  
- Booklore version (e.g., v0.35.0): 1.2.1
- What computer or device are you using? (Windows, Mac, Linux, etc.): Proxmox VE LXC to host, and browser On Windows 11 to access
- Which browser and version? (Chrome, Firefox, Safari, etc.): Chrome 140.0.7339.128
- How did you install Booklore? (Docker, manual install, etc.): Proxmox community script, [here](https://community-scripts.github.io/ProxmoxVE/scripts?id=booklore)

PR #1459, opened 2025-10-27
fix: tie book uploading topbar item to upload book permission

heyitsbench:
This closes https://github.com/booklore-app/booklore/issues/1141. Tested in a Docker container with:
- [An admin account (full permissions)](https://github.com/user-attachments/assets/b6ba7e72-e88d-4108-b0c2-3ebce776ec95)
- [A user with all permissions except Admin and Upload Book](https://github.com/user-attachments/assets/bf295e97-612e-4baf-b2c8-9dab24cd11e4)
- [A user with only Upload Book permissions](https://github.com/user-attachments/assets/ad26afb0-ee52-4bc2-9c6b-0b98be3067ef).

I did notice an issue with a gap between items when permissions are not high enough, but given that this happens on the base repository and is not introduced per se with this PR, I've elected not to chase it.

### DIFF
--- a/booklore-ui/src/app/shared/layout/component/layout-topbar/app.topbar.component.html
+++ b/booklore-ui/src/app/shared/layout/component/layout-topbar/app.topbar.component.html
@@ -45,7 +45,7 @@
             }
           </li>
           <li>
-            @if (userState.user?.permissions?.canManipulateLibrary || userState.user?.permissions?.admin) {
+            @if (userState.user?.permissions?.canUpload || userState.user?.permissions?.admin) {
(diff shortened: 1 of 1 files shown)


heyitsbench:
Unrelated to this PR, but I did have an issue with [app.routes.tsL13](https://github.com/booklore-app/booklore/blob/develop/booklore-ui/src/app/app.routes.ts#L13), my Docker deployment failed due to it not being able to find the [file mentioned](https://github.com/booklore-app/booklore/blob/develop/booklore-ui/src/app/shared/components/setup/%20login.guard.ts). I fixed this locally by removing the space in the path, but I'm not sure if that's a correct fix or if it's hitting any other users.

PR #1486, opened 2025-10-30
Enhancement: Import custom pagecount field from calibre

beedaddy:
Calibre exports custom columns as JSON strings. This PR parses this JSON data and adds the pagecount value to booklore. By convention, the calibre column must be named "pagecount".

### DIFF
--- a/booklore-api/build.gradle
+++ b/booklore-api/build.gradle
@@ -65,7 +65,6 @@ dependencies {
     // --- JSON & Web Scraping ---
     implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.20.0'
     implementation 'org.jsoup:jsoup:1.21.2'
-    implementation 'org.json:json:20250517'
 
     // --- Mapping (DTOs & Entities) ---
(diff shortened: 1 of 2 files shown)


booklore-api/build.gradle
```diff
@@ -65,6 +65,7 @@ dependencies {
     // --- JSON & Web Scraping ---
     implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.20.0'
     implementation 'org.jsoup:jsoup:1.21.2'
+    implementation 'org.json:json:20250517'
```
acx10: Could you try implementing this without adding an external library? I believe it can be done using Jackson or another library that’s already included in the project.

adityachandelgit (changes requested):

booklore-api/build.gradle
beedaddy: Of course, sorry for that.

beedaddy:
It would be best to mention somewhere that the field (the Calibre column) must be called “pagecount.” Alternatively—but I don't know enough about this—you could also make the name configurable.

Issue #555, opened 2025-06-23
[feature request] - Share shelves with users


mobiledude:
Installed it, and got OIDC up and running with Authentik. Really nice tool!

Would love to have the ability to "share" a Shelve amongst users. ( e.g. having the option 2 different users can manage a shelf )

To start of easily, creating "public"  vs "private" shelves could solve this

adityachandelgit:
You’re right, currently, Booklore only supports sharing at the library level. Shelf sharing is a great idea and would add flexibility. I’ll look into whether the current DB structure can support this cleanly and consider it for a future release.

devra1:
+1. This would be really useful. TIA.

adityachandelgit:
Closing this since it’s not in scope, sorry!

cjd:
I was looking at this the other day and I think it can be done by adding a boolean such as is_public to the magic_shelf table and locking down changes to shelves marked public to only admin users.
@adityachandelgit If I worked on implementing this (even just for magic shelves) - would that be okay or is it too far out of scope?

PR #1489, opened 2025-10-31
Add ability to mark magic shelves as public

cjd:
Adds a checkbox to the create/edit magic shelf dialog to set the shelf as public.
If public then it shows up for all users.
Only admins can edit public shelves (edit options disabled for non-admin)

Admin view:
2025-10-31T12:38:19,199542957+11:00

Non-admin view:
2025-10-31T12:34:01,768008180+11:00

Implements FR #555

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/config/security/JwtUtils.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/config/security/JwtUtils.java
@@ -15,6 +15,7 @@ import org.springframework.stereotype.Service;
 
 import javax.crypto.SecretKey;
 import java.nio.charset.StandardCharsets;
+import java.time.Instant;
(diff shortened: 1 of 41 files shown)


cjd:
If this is suitable, I can probably extend to work in the same way for normal shelves as well

acx10:
@cjd Hi, this looks great at a glance, nicely done. I’m currently on vacation, but I’ll run and test the code once I’m back and get it merged soon after.

And yes, we can definitely handle regular shelves in a similar way. Thanks for the great work!

thob:
hey, I'm observing an odd behavior of said public shelf, not all users are equal it seems, posted it over at - hope that'S ok.  > https://www.reddit.com/r/BookLoreApp/comments/1q9vee3/magic_shelf_visibility_public_which_perms_required/

PR #1538, opened 2025-11-09
Feature: Add EPUB conversion support for CBX files to Kobo Sync

CounterClops:
Adds support for syncing CBX files on Kobo through the in-built Kobo sync. Using the images from a CBX file, it generates a EPUB including all the required content/metadata files. It currently extracts all images to disk before repackaging them into the EPUB file, so it shouldn't have the memory issues from the previous code.

The CBX to EPUB conversion is disabled by default, and CBX files will not show up in the Kobo sync unless the CBX to EPUB conversion is enabled. There is also a CBX file size limit, to avoid listing CBX files in the sync when they're too large for the conversion service.

This also does not do conversion chaining, so if you have the EPUB -> KEPUB configured, it won't do CBX -> EPUB -> KEPUB conversions. They both function separately, and KEPUB conversion will only trigger on EPUBS that are original EPUBS, not converted from CBX.

I've generated some unit tests, which I have used to validate the CBX conversion.

This has also been directly tested using my own Kobo Libra Colour, and I've confirmed both CBZ, and CBR files sync across as expected.

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/config/security/service/OpdsUserDetailsService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/config/security/service/OpdsUserDetailsService.java
@@ -24,6 +24,6 @@ public class OpdsUserDetailsService implements UserDetailsService {
         OpdsUserV2Entity userV2 = opdsUserV2Repository.findByUsername(username)
                 .orElseThrow(() -> ApiError.USER_NOT_FOUND.createException(username));
         OpdsUserV2 mappedCredential = opdsUserV2Mapper.toDto(userV2);
-        return new OpdsUserDetails(mappedCredential, null);
+        return new OpdsUserDetails(mappedCredential);
     }
 }
--- a/booklore-api/src/main/java/com/adityachandel/booklore/config/security/userdetails/OpdsUserDetails.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/config/security/userdetails/OpdsUserDetails.java
@@ -1,6 +1,5 @@
 package com.adityachandel.booklore.config.security.userdetails;
 
-import com.adityachandel.booklore.model.dto.OpdsUser;
 import com.adityachandel.booklore.model.dto.OpdsUserV2;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
@@ -15,20 +14,15 @@ import java.util.List;
 public class OpdsUserDetails implements UserDetails {
 
     private final OpdsUserV2 opdsUserV2;
-    private final OpdsUser opdsUser;
 
     @Override
     public String getUsername() {
-        if (opdsUserV2 != null) return opdsUserV2.getUsername();
-        if (opdsUser != null) return opdsUser.getUsername();
-        throw new IllegalStateException("No username available");
+        return opdsUserV2.getUsername();
     }
 
     @Override
     public String getPassword() {
-        if (opdsUserV2 != null) return opdsUserV2.getPasswordHash();
-        if (opdsUser != null) return opdsUser.getPassword();
-        throw new IllegalStateException("No password available");
+        return opdsUserV2.getPasswordHash();
     }
 
     @Override
--- a/booklore-api/src/main/java/com/adityachandel/booklore/controller/OpdsController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/controller/OpdsController.java
@@ -38,8 +38,7 @@ public class OpdsController {
         @ApiResponse(responseCode = "404", description = "Book not found")
     })
     @GetMapping("/{bookId}/download")
-    public ResponseEntity<Resource> downloadBook(
-            @Parameter(description = "ID of the book to download") @PathVariable("bookId") Long bookId) {
+    public ResponseEntity<Resource> downloadBook(@Parameter(description = "ID of the book to download") @PathVariable("bookId") Long bookId) {
         return bookService.downloadBook(bookId);
     }
 
@@ -49,8 +48,7 @@ public class OpdsController {
         @ApiResponse(responseCode = "404", description = "Book or cover not found")
     })
     @GetMapping("/{bookId}/cover")
-    public ResponseEntity<Resource> getBookCover(
-            @Parameter(description = "ID of the book") @PathVariable long bookId) {
+    public ResponseEntity<Resource> getBookCover(@Parameter(description = "ID of the book") @PathVariable long bookId) {
         Resource coverImage = bookService.getBookThumbnail(bookId);
         String contentType = "image/jpeg";
         return ResponseEntity.ok()
@@ -62,8 +60,7 @@ public class OpdsController {
     @Operation(summary = "Get OPDS root catalog", description = "Retrieve the OPDS root navigation feed.")
     @ApiResponse(responseCode = "200", description = "Root OPDS catalog returned successfully")
     @GetMapping(produces = OPDS_CATALOG_MEDIA_TYPE)
-    public ResponseEntity<String> getRootCatalog(
-            @Parameter(hidden = true) HttpServletRequest request) {
+    public ResponseEntity<String> getRootCatalog(@Parameter(hidden = true) HttpServletRequest request) {
         String feed = opdsFeedService.generateRootNavigation(request);
         return ResponseEntity.ok()
                 .contentType(MediaType.parseMediaType(OPDS_CATALOG_MEDIA_TYPE))
@@ -73,8 +70,7 @@ public class OpdsController {
     @Operation(summary = "Get OPDS libraries navigation", description = "Retrieve the OPDS libraries navigation feed.")
     @ApiResponse(responseCode = "200", description = "Libraries navigation feed returned successfully")
     @GetMapping(value = "/libraries", produces = OPDS_CATALOG_MEDIA_TYPE)
-    public ResponseEntity<String> getLibrariesNavigation(
-            @Parameter(hidden = true) HttpServletRequest request) {
+    public ResponseEntity<String> getLibrariesNavigation(@Parameter(hidden = true) HttpServletRequest request) {
         String feed = opdsFeedService.generateLibrariesNavigation(request);
         return ResponseEntity.ok()
                 .contentType(MediaType.parseMediaType(OPDS_CATALOG_MEDIA_TYPE))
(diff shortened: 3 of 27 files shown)


booklore-api/src/main/java/com/adityachandel/booklore/service/book/BookDownloadService.java
```diff
@@ -74,24 +76,32 @@ public ResponseEntity<Resource> downloadBook(Long bookId) {
...
 
         Path tempDir = null;
         try {
             File inputFile = new File(FileUtils.getBookFullPath(bookEntity));
             File fileToSend = inputFile;
 
+            if (convertCbx) {
+                tempDir = Files.createTempDirectory("cbx-epub-output");
+                fileToSend = cbxConversionService.convertCbxToEpub(inputFile, tempDir.toFile(), bookEntity);
+            }
```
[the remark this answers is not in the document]
CounterClops: This is being handled in the finally block at the end of the function, this means if a error occurs anywhere whether it's in the EPUB to KEPUB conversion or CBX to EPUB conversion, it'll do the temp folder cleanup.

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
```diff
@@ -0,0 +1,542 @@
...
+            addImageToZip(zipOut, imagePath, image);
+
+            String htmlContent = generatePageHtml(imageFileName, i + 1);
+            ZipArchiveEntry htmlEntry = new ZipArchiveEntry(htmlPath);
+            zipOut.putArchiveEntry(htmlEntry);
+            zipOut.write(htmlContent.getBytes(StandardCharsets.UTF_8));
+            zipOut.closeArchiveEntry();
+
+            contentGroups.add(new EpubContentFileGroup(contentKey, imagePath, htmlPath));
+        }
```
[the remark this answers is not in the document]
CounterClops: I want to keep the cover image as a seperate file in the epub file, so if we ever make any changes like make the cover use a custom cover image from booklore, or modify the cover in some other fashion, it won't affect page 1 it won't cause issues.

booklore-api/src/test/java/com/adityachandel/booklore/service/kobo/KoboCompatibilityServiceTest.java
```diff
@@ -0,0 +1,296 @@
...
+        koboSettings.setConvertCbxToEpub(true);
+        koboSettings.setConversionLimitInMbForCbx(50);
+        when(appSettingService.getAppSettings()).thenReturn(appSettings);
+
+        boolean isSupported = koboCompatibilityService.isBookSupportedForKobo(cbxBookAtLimit);
+
+
+        assertThat(isSupported).isTrue();
+        verify(appSettingService, atLeastOnce()).getAppSettings();
+    }
```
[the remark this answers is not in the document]
CounterClops: The display name makes this explicit, and so does the function test name? It's not hard to add a comment for this, but I feel like it's already covered well enough?

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
```diff
@@ -0,0 +1,569 @@
...
+                cbxFile.getName().replaceFirst("\\.[^.]+$", "") + ".epub");
+        File epubFile = epubFilePath.toFile();
+
+        List<BufferedImage> images = extractImagesFromCbx(cbxFile);
+        if (images.isEmpty()) {
+            throw new IllegalStateException("No valid images found in CBX file: " + cbxFile.getName());
+        }
+
+        log.debug("Extracted {} images from CBX file", images.size());
+
```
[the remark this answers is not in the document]
CounterClops: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/booklore-app/booklore/pull/1538#discussion_r2554015996)

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
```diff
@@ -0,0 +1,569 @@
...
+    private String makeRelativeToOebps(String fullPath) {
+        Path oebpsPath = Paths.get("OEBPS");
+        Path targetPath = Paths.get(fullPath);
+        
+        if (targetPath.startsWith(oebpsPath)) {
+            return oebpsPath.relativize(targetPath).toString();
+        }
+        
+        return fullPath;
+    }
```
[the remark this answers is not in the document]
CounterClops: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/booklore-app/booklore/pull/1538#discussion_r2554015995)

booklore-api/src/main/java/com/adityachandel/booklore/service/book/BookDownloadService.java
```diff
@@ -77,26 +79,36 @@ public ResponseEntity<Resource> downloadBook(Long bookId) {
...
+                tempDir = Files.createTempDirectory("kobo-conversion");
+            }
+
+            if (convertCbx) {
+                fileToSend = cbxConversionService.convertCbxToEpub(inputFile, tempDir.toFile(), bookEntity);
+            }
+
             if (asKepub) {
-                tempDir = Files.createTempDirectory("kepub-output");
                 fileToSend = kepubConversionService.convertEpubToKepub(inputFile, tempDir.toFile());
```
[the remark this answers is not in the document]
CounterClops: This is intentional, as the KEPUB conversion should not happen for CBX files that have been converted to EPUB. The check for whether a file should be converted to KEPUB happens before the conversion, and requires the source file is a EPUB, it can't be a CBX file to start with.

We could chain the conversion, but I'm not sure the extra KEPUB features would be useful for comics, and I'm unsure what changes would be made to the EPUBs to convert them to KEPUBs and whether they would cause any issues with viewing.

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
```diff
@@ -0,0 +1,635 @@
...
+
+    private File executeCbxConversion(File cbxFile, File tempDir, BookEntity bookEntity) 
+            throws IOException, TemplateException, RarException {
+        
+        Path epubFilePath = Paths.get(tempDir.getAbsolutePath(),
+                cbxFile.getName().replaceFirst("\\.[^.]+$", "") + ".epub");
+        File epubFile = epubFilePath.toFile();
+
+        List<BufferedImage> images = extractImagesFromCbx(cbxFile);
```
balazs-szucs: Doesn't this straight up loads _all images_ into memory at once? I haven't taken a very close look, but it feels like this should wrapped in a try with InputStream...catch, if possible.

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
```diff
@@ -0,0 +1,635 @@
...
+                cbxFile.getName(), outputFile.getName(), outputFile.length());
+        return outputFile;
+    }
+
+    private File executeCbxConversion(File cbxFile, File tempDir, BookEntity bookEntity) 
+            throws IOException, TemplateException, RarException {
+        
+        Path epubFilePath = Paths.get(tempDir.getAbsolutePath(),
+                cbxFile.getName().replaceFirst("\\.[^.]+$", "") + ".epub");
+        File epubFile = epubFilePath.toFile();
```
balazs-szucs, line 122: This also not sanitized or?

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
```diff
@@ -0,0 +1,635 @@
...
+
+        List<BufferedImage> images = extractImagesFromCbx(cbxFile);
+        if (images.isEmpty()) {
+            throw new IllegalStateException("No valid images found in CBX file: " + cbxFile.getName());
+        }
+
+        if (images.size() > WARNING_IMAGE_COUNT_THRESHOLD) {
+            log.warn("CBX file contains {} images, which may cause memory issues on systems with limited heap space", images.size());
```
balazs-szucs: Warning about size AFTER loading everything into memory seem a bit redundant, either do before and let user decide if this is they want or just do it :sweat:

balazs-szucs: Sorry, for the spam, I didn't realise I wasn't in the "review" mode, so if you got notification for every edit/delete/re-make, my apologizes.

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
CounterClops: This does load into memory at once, I went with this approach as it simplifies the logic between CBZ, CBR, and CB7 files as I believe they'll have different streaming characteristics and I wanted the logic that pulls the images out to be separate from the core logic around making the epub.

I could extract all files to disk instead, then load them into the EPUB creation logic, but given this is a on the fly conversion I want it to be fast so the sync doesn't get held up. Might still be fine though, so worth testing

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
CounterClops: Prior to even reaching this point, it has a check to limit CBX file size. So it won't process files above that size threshold, this can be adjusted by the booklore admin, but by default it won't convert and CBX files above 100MB. Admins can lower or raise this based on how resource constrained their booklore instance is.

It's likely a bit redundant to have the log after, but it's more just a warning. Could probably remove it.

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
balazs-szucs: Thanks for the reply. Generally I am a much bigger fan of the disk (into tempDir) approach. I feel like somewhere on the host OS, there should be some buffering/optimization happening under the hood, so I don't personally believe that'd be so much slower than this, but it also comes with the benefit of lower OOM risk. (Speed generally depends on the available RAM, I'd wager that if RAM is low, writing to disk would actually be faster.)

CounterClops: To be honest I had been debating coming back and switching it to disk instead, so I'll change it up. Shouldn't be too hard since most the logic for getting the images is separate to the construction of the EPUB.

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
CounterClops: I can probably work this without a regex, but it's just taking the filename from the CBX file and replacing it with EPUB. I might just take the KEPUB approach and just append the .EPUB to the end so it'll end up being something like filename.cbz.epub

booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/CbxConversionService.java
CounterClops: Switched it up so it now extracts to the disk first, it is a fair bit slower in my testing that reading into memory, but not unusable. Would like to test it on syncing/downloading larger libraries. 

But should avoid the out of memory issues more safely, so I think it's a better method overall.

CounterClops:
@adityachandelgit Anything more you need from me on this one? It now extracts images to disk instead of memory for the CBX to EPUB conversion, and I've gone through all the issues raised by the Copilot review that seemed legitimate, and made changes.

acx10:
@CounterClops. At a glance everything looks good. I’ll review the code more thoroughly and proceed with the merge accordingly. Thanks!

Issue #1521, opened 2025-11-04
I believe the file move refactor in 1.9 broke file moving


goaliedude3919:
# Bug Report Template for Booklore

**What happened?**  
After upgrading from 1.4 to 1.10, Organize File no longer works.

**How can we see it happen?**  
Steps to reproduce the issue:  

1. Open Booklore and go to a book info dialog.
2. Select the "Three dots menu" and select "Organize Files".
3. Then select Move Files from the "Organize Book Files" dialog.
4. It says that the "file successfully moved" but looking at the actual folder, it's still in the original location with the original filename.

**What did you expect to happen?**  
The book files should be properly renamed/moved.

**Screenshots or Error Messages**  
This is the only log message when initiating a file rename: `2025-11-04T16:47:27.051-05:00  INFO 10 --- [booklore-api] [io-8080-exec-12] c.a.b.s.monitoring.MonitoringService     : Registered 131 folders for library 'Books'`

**About your setup:**  
- Booklore version v1.10.0
- What computer or device are you using? Beelink S12 for the Booklore container, Synology NAS for the media storage 
- Which browser and version? Edge 141.0.3537.99 
- How did you install Booklore? Docker on Debian 12

**Anything else we should know?**  
I believe the issue is likely related to this enhancement in 1.9: https://github.com/booklore-app/booklore/pull/1406

ajb3932:
Same issue for me, I thought it was a file permission error but still no luck

PR #1582, opened 2025-11-19
Fix file moving - fixes issue #1521

cjd:
This fixes file moving issues seen in bugs #1521 and #1440

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/util/MetadataChangeDetector.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/util/MetadataChangeDetector.java
@@ -112,10 +112,19 @@ public class MetadataChangeDetector {
         compareValue(diffs, "comicvineId", clear.isComicvineId(), newMeta.getComicvineId(), existingMeta.getComicvineId(), () -> !isTrue(existingMeta.getComicvineIdLocked()));
         compareValue(diffs, "hardcoverId", clear.isHardcoverId(), newMeta.getHardcoverId(), existingMeta.getHardcoverId(), () -> !isTrue(existingMeta.getHardcoverIdLocked()));
         compareValue(diffs, "googleId", clear.isGoogleId(), newMeta.getGoogleId(), existingMeta.getGoogleId(), () -> !isTrue(existingMeta.getGoogleIdLocked()));
+        compareValue(diffs, "pageCount", clear.isPageCount(), newMeta.getPageCount(), existingMeta.getPageCount(), () -> !isTrue(existingMeta.getPageCountLocked()));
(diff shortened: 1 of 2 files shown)


booklore-api/src/main/java/com/adityachandel/booklore/util/MetadataChangeDetector.java
```diff
@@ -112,10 +112,19 @@ public static boolean hasValueChangesForFileWrite(BookMetadata newMeta, BookMeta
...
+        compareValue(diffs, "hardcoverRating", clear.isHardcoverRating(), newMeta.getHardcoverRating(), existingMeta.getHardcoverRating(), () -> !isTrue(existingMeta.getHardcoverRatingLocked()));
+        compareValue(diffs, "hardcoverReviewCount", clear.isHardcoverReviewCount(), newMeta.getHardcoverReviewCount(), existingMeta.getHardcoverReviewCount(), () -> !isTrue(existingMeta.getHardcoverReviewCountLocked()));
         compareValue(diffs, "authors", clear.isAuthors(), newMeta.getAuthors(), toNameSet(existingMeta.getAuthors()), () -> !isTrue(existingMeta.getAuthorsLocked()));
         compareValue(diffs, "categories", clear.isCategories(), newMeta.getCategories(), toNameSet(existingMeta.getCategories()), () -> !isTrue(existingMeta.getCategoriesLocked()));
+        compareValue(diffs, "moods", clear.isMoods(), newMeta.getMoods(), toNameSet(existingMeta.getMoods()), () -> !isTrue(existingMeta.getMoodsLocked()));
+        compareValue(diffs, "tags", clear.isTags(), newMeta.getTags(), toNameSet(existingMeta.getTags()), () -> !isTrue(existingMeta.getTagsLocked()));
```
acx10, line 127: IMO, fields like amazonRating, amazonReviewCount, goodreadsRating, hardcoverRating, etc. should not be written to the metadata file. These values are dynamic and change often, so they should remain Booklore-only fields stored in the DB, not persisted back to the file itself.

Can you clarify why we want to write these ephemeral ratings/review counts into the file? From a data-design perspective, it feels like only stable metadata (IDs, authors, categories, page count, etc.) should go into the file, while volatile fields stay in Booklore.

booklore-api/src/main/java/com/adityachandel/booklore/service/file/FileMoveService.java
```diff
@@ -66,10 +66,7 @@ public void bulkMoveFiles(FileMoveRequest request) {
                 LibraryPathEntity libraryPathEntity = optionalLibraryPathEntity.get();
 
                 LibraryEntity sourceLibrary = bookEntity.getLibrary();
-                if (sourceLibrary.getId().equals(targetLibrary.getId())) {
```
acx10, line 69: Ahh, so this was the cause of the bug, thanks for fixing.

Issue #1554, opened 2025-11-13
Feature Request: Typeahead for Series Name


Na3blis:
**What problem or limitation are you encountering?**  
Have to go back to the library to look at what a series is named, mainly around if it has `The` or something similar, such as `The Dresden files` vs `Dresden Files`. 

**What solution or improvement do you propose?**  
Several fields in the metadata have drop down type-ahead, would be good to add that to the series name as well

**Have you found any workarounds or alternatives?**  
Just to go back and forth between library and editing metadata

PR #1587, opened 2025-11-20
Feature: Autocomplete for publisher and series on book metadata edit screen

Muppetteer:
As requested here: #1554


Muppetteer:
I added autocomplete to the bulk edit form. Also moved a couple of fields so the order made more sense and reduced the size of the form.

Issue #1572, opened 2025-11-18
Metadata is not being written to epub files, lost everything


Muppetteer:
**What happened?**  
I tried creating a new library and accidentally used the same path as another existing library (there should be a warning to stop this from happening btw). This duplicated all the books in my libraries collection. So I deleted the new library and re-scanned the old one. After this every single edit made to the metadata of my books was lost apart from updated cover images, despite the option to write metadata to file being set. Hundreds of hand-edits lost just like that and I have to start all over again.

I was annoyed the database lost everything I had edited but thought no problem because the metadata is written to the files and can be re-scanned. But no. So I did an experiment with a single book. I downloaded a copy then edited metadata and downloaded it again. The second download has no edits to the metadata when I look through the epub source files and the files have an identical number of bytes and SHA hash.

**How can we see it happen?**  
Steps to reproduce the issue:  
1. Download a book and keep it safe.
2. Ensure write metadata to file option is set.
3. Edit the metadata of the book, add some genres, scores etc.
4. Download the book again and compare the epub source files.

**What did you expect to happen?**  
Edited metadata should be in the updated book file.

**Screenshots or Error Messages**  
SHA of original book:
14c2771070d19487b74f7af2748cc5eab16edc92e399487185802ac45550e488  ./book - original.epub

SHA of edited book:
14c2771070d19487b74f7af2748cc5eab16edc92e399487185802ac45550e488  ./book - changed metadata.epub

I can also see on the server that the file date does not change on the epub after editing metadata.

**About your setup:**  
- Booklore version 1.10
- Linux NAS box, all books on same system as Booklore (no network FS)
- Docker install

Muppetteer:
Possibly related #1562

cjd:
The problem was that MetadataChangeDetector.hasValueChangesForFileWrite was being called _after_ the metadata was changed (so it was comparing new data to new data - and thinking there was nothing to be written)
I've just made a pull request https://github.com/booklore-app/booklore/pull/1592 to fix this up

PR #1592, opened 2025-11-21
Fix metadata writing to file

cjd:
Moves the check for metadata changes to a bit earlier so it is comparing to pre-update data.

Fixes bug #1572


dcogny:
This is great, thanks, that said, it would be good to have some way to force bulk "Write Metadata to Files" without having to make changes, I have a lot of books that I have cleaned up the metadata but it was never added to the file and now, I don't know which not I have nay metadata to update.
I am thinking about adding a "fixing" tag to all of them and then removing the tag, that could be a workaround and I leave it here for others interested, but still, it would be nice to have the option on a library or shelved to "Force Metadata Write".

Issue #1637, opened 2025-11-25
Files deleted from booksdrop on failed import


ckozak:
# 🐛 Bug Report for Booklore

Thank you for helping us improve Booklore! Please fill out the sections below.

---

## 📝 What happened?
When I import files via booksdrop folder, some files in a batch of 5+ books usually fail to import. Aside of the actual import issue (usually duplicate not allowed or batch issue), the file is deleted from booksdrop folder, although "importer" still thinks the files are there, but it fails on retry. 

That's a serious data issue because users have to manually look at what files were failed, and then re-copy them into the booksdrop folder. It's even much larger issue when booksdrop folder is auto-populated by downloaders, so the file needs to be re-downloaded

## 🔄 Steps to Reproduce

1. Drop 5-10 files into booksdrop folder (v 1.11 and 1.12 have the issue)
2. Wait for booksdrop scanner to process the books
3. Select all books and press "finalize"
4. Some files will fail, and the source data files will be deleted from booksdrop

## ✅ Expected Behavior
Failed files should be preserved, to allow for retries and troubleshooting

## 💻 Environment
- **Version:** 1.11, 1.12
- **OS:** - Docker container running on Ubuntu and on Synology
- **Browser:** Safari and Firefox
- **Installation:** - Docker

## ✨ Possible Solution _(Optional)_
Haven't seen the code, but files should be deleted from booksdrop only on success. Or, at least, there should be property to control that behaviour

Issue #1537, opened 2025-11-08
[BUG] Errors when adding more than one file via bookdrop


eXaminator:
# Bug Report Template for Booklore

**What happened?**  
When I have books in my bookdrop and I try to finalize multiple books while also updating their metadata, I get an SQL error for some of them (see screenshot). The books with errors stay in the bookdrop list, but their files are removed from the filesystem. 

**How can we see it happen?**  
Steps to reproduce the issue:  
1. Add at least 2 files to the bookdrop folder
2. Open Booklore and go to the bookdrop list
3. Apply the metadata from the automatic book matching
4. The error in the screenshot appears.

**What did you expect to happen?**  
No errors should appear and the books should be added to my library with the selected metadata.

**Screenshots or Error Messages**  
<img width="1962" height="192" alt="Image" src="https://github.com/user-attachments/assets/23c2a190-be1c-4c9d-9bfb-55d07bbfcf02" />

<img width="714" height="524" alt="Image" src="https://github.com/user-attachments/assets/5790b912-10ce-44dd-819f-74b04f55f610" />

**About your setup:**  
- Booklore version: 1.10.0
- What computer or device are you using? Windows PC
- Which browser and version? Zen (based on Firefox) 1.17.6b
- How did you install Booklore? Docker on my Zimaboard 2 Server (x86)

**Anything else we should know?**  
I have enabled metadata updates for files. The files are local on my server (which is also my NAS). The import works well when I add files to my library one by one.

It also works if I JUST finalize multiple files without updating their Metadata.

Edit: Ok, it works better when not editing metadata, but I just got the errot while finalizing 5 books without updating their metadata first (so just selecting them as is and then clicking finalize).

PR #1638, opened 2025-11-25
fix(bookdrop): Ensure source file is only deleted after successful import and cleanup target on failure

balazs-szucs:
Improved the reliability of file handling in the bookdrop import process and enhances the reliability of related tests. The main change ensures that source files are only deleted after successful processing, and that target files are cleaned up if processing fails, preventing orphaned or lost files. The tests are updated to verify these behaviors and to use lenient mocking for file system operations, making them more resilient and easier to maintain.

**File move and cleanup logic improvements:**

* Updated `performFileMove` in `BookDropService.java` to delete the source file only after successful processing, and to clean up the target file if processing fails or returns a failure result. This prevents accidental loss of files and ensures proper cleanup in error scenarios.

**Testing reliability and coverage:**

* Added three new tests to `BookDropServiceTest.java` to verify:
  - Source file is deleted only on successful processing.
  - Source file is preserved and target file is cleaned up when processing fails.
  - Source file is preserved and target file is cleaned up when metadata refresh fails.

**Test mocking improvements:**

* Changed all usages of `mockStatic(Files.class)` in `BookDropServiceTest.java` to use lenient settings, reducing brittleness and allowing for more flexible test setups.
* Enabled lenient strictness globally for the test class with `@MockitoSettings(strictness = Strictness.LENIENT)`.

Closes: #1637

### DIFF
--- a/booklore-api/src/test/java/com/adityachandel/booklore/service/bookdrop/BookDropServiceTest.java
+++ b/booklore-api/src/test/java/com/adityachandel/booklore/service/bookdrop/BookDropServiceTest.java
@@ -15,6 +15,7 @@ import com.adityachandel.booklore.model.entity.LibraryEntity;
 import com.adityachandel.booklore.model.entity.LibraryPathEntity;
 import com.adityachandel.booklore.repository.BookRepository;
 import com.adityachandel.booklore.repository.BookdropFileRepository;
+import com.adityachandel.booklore.repository.LibraryPathRepository;
 import com.adityachandel.booklore.repository.LibraryRepository;
 import com.adityachandel.booklore.service.NotificationService;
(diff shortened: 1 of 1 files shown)


Muppetteer:
This might fix #1537 too, at least the failed import books won't be lost.

balazs-szucs:
Yeah, I agree, it's a partial fix. I slept on it, and I think I can look into the other issues later (and keep this as-is, for easier review), like throwing better SQL exceptions and addressing possible race conditions during upload. But for now, this is good enough and addresses the main issue. I'd like to handle the remaining stuff in separate PRs. Probably might even separate the two, but we'll see.

Thanks for your comment :)

PR #1662, opened 2025-11-28
Enable marking Shelves as 'public'

cjd:
Refactor a few things so shelves can be shared
Creation/editing of public shelves is limited to admins


cjd:
I have tested this a fair bit myself - and think I've caught the various use cases (adding books to shelves now confirms you are allowed to add to a shelf) but I may have missed some corner case that I didn't think of.
If anyone notices a breakage I missed - let me know

acx10:
I’m running into a strange issue where the books in my libraries aren’t showing up in the UI. Here’s the video recording of what I’m seeing:

https://github.com/user-attachments/assets/fd104a9c-89e7-4f7b-8e9a-31b712cb9b7b

PR #1671, opened 2025-11-29
Add single select filter mode

### LINKED ISSUE #1640: [Feature]: add None/Alternate as filter method
# Feature Request Template for Booklore

**What problem or limitation are you encountering?**  
The Filter mode works a little counter-intuitive to my liking.
I'm using it as a selection method and sometimes am just walking through all the authors or series.
So when I click an author and want to go to another author, I first have to deselect the previous one and then click the next one. It's annoying, because many times I forget to do that and end up with an empty selection.

**What solution or improvement do you propose?**  
Instead of only And and Or as a filter method, add a None (or something similar named lik Alternate) and make the default Filter method adjustable in the view settings. Adjustable, because other uses may just like this style of filtering.

**Have you found any workarounds or alternatives?**  

**Additional details**

Muppetteer:
Fixes #1640

### DIFF
--- a/booklore-ui/src/app/features/book/components/book-browser/book-filter/book-filter.component.ts
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-filter/book-filter.component.ts
@@ -169,7 +169,7 @@ export class BookFilterComponent implements OnInit, OnDestroy {
   truncatedFilters: Record<string, boolean> = {};
   filterTypes: string[] = [];
   filterModeOptions = [
-    {label: '&', value: 'and'},
+    {label: 'AND', value: 'and'},
     {label: 'OR', value: 'or'},
     {label: '1', value: 'single'},
   ];


booklore-ui/src/app/features/book/components/book-browser/book-filter/book-filter.component.ts
```diff
@@ -167,10 +169,12 @@ export class BookFilterComponent implements OnInit, OnDestroy {
   truncatedFilters: Record<string, boolean> = {};
   filterTypes: string[] = [];
   filterModeOptions = [
-    {label: 'AND', value: 'and'},
-    {label: 'OR', value: 'or'}
+    {label: '&', value: 'and'},
```
acx10:
```suggestion
    {label: 'AND', value: 'and'},
```

acx10:
Rest of it looks good!

acx10:
You can also try overriding the padding for the toggle button to make it more compact:

```
--p-togglebutton-content-padding: 0.25rem 0.5rem;
```

Screenshot 2025-11-28 at 6 34 50 PM

Muppetteer:
I had "AND" initially but the buttons would not fit. I later changed the text label to the left and didn't check again. Looks like it just fits.

filters

Issue #1404, opened 2025-10-21
Serve Angular app directly from the Java backend


wariuccio:
Hello,
I was looking at the Dockerfile and at the configuration and I noticed that Nginx there is adding some complexity. Admittedly not too much but not zero.

I am wondering if Nginx is actually required or if the Angular app could be served directly by the Java backend. This would remove Nginx running from inside the Docker container (which is a win). Usually I like to have a single process in each container for ease of maintenance.

Personally when I use Docker containers I already have a reverse proxy in front of all of them that does a simple `proxyPass` to whatever the container is doing. I can still do it with BookLore but that means that I have now 2 reverse proxies running, my own and the ones inside the Docker container.

If possible, I think this chance would be a win by reducing both the configuration and operational complexity of running BookLore.

adityachandelgit:
I’m interested in implementing this change. Would you be willing to collaborate on it?

wariuccio:
I can try to have a look but I'm not familiar with Angular nor with Spring Boot. Good reason to learn something new :)

Should I create a PR once I am ready (it might take a while) or were you thinking to collaborate in some different way?

PR #1679, opened 2025-11-29
Serve the Angular app from Spring boot

wariuccio:
Addresses https://github.com/booklore-app/booklore/issues/1404

This PR updates the Spring Boot application to make it able to serve the pre-built files of the Angular app from a location on disk. This in turn enables us to get rid of Nginx since we only have a single web server (the Spring Boot application) able to serve everything we need.

I tested this with the example Docker compose file (substituting in a Docker image I built myself with the changes in this PR) and it seems to work.
There are a couple things I am not too sure about, I'll leave comments inline.


booklore-api/src/main/java/com/adityachandel/booklore/config/WebMvcConfig.java
```diff
@@ -1,20 +1,49 @@
...
     private final OpdsEnabledInterceptor opdsEnabledInterceptor;
 
     @Override
     public void addInterceptors(InterceptorRegistry registry) {
         registry.addInterceptor(opdsEnabledInterceptor)
                 .addPathPatterns("/api/v1/opds/**", "/api/v2/opds/**");
     }
+
+    @Override
+    public void addViewControllers(ViewControllerRegistry registry) {
```
wariuccio, line 29: This is required to redirect requests like `localhost:8080` or `localhost:8080/`

booklore-api/src/main/java/com/adityachandel/booklore/config/WebMvcConfig.java
```diff
@@ -1,20 +1,49 @@
...
+        registry.addViewController("/").setViewName("forward:/index.html");
+    }
+
+    @Override
+    public void addResourceHandlers(ResourceHandlerRegistry registry) {
+        String angularPath = appProperties.getAngularPath();
+        registry.addResourceHandler("/**")
+                .addResourceLocations("file:"+angularPath)
+                .resourceChain(true)
+                .addResolver(new PathResourceResolver() {
```
wariuccio, line 39: Custom path resolver is required to have Angular routing work

booklore-api/src/main/java/com/adityachandel/booklore/config/security/SecurityConfig.java
```diff
@@ -139,6 +139,7 @@ public SecurityFilterChain jwtApiSecurityChain(HttpSecurity http) throws Excepti
             publicEndpoints.addAll(Arrays.asList(SWAGGER_ENDPOINTS));
         }
         http
+                .securityMatcher("/api/**")
```
wariuccio, line 142: We don't want this to math on everything as the Angular app doesn't need to be protected. Should also this match `/ws/**`?

booklore-api/src/main/resources/application.yaml
```diff
@@ -17,6 +18,8 @@ app:
 server:
   forward-headers-strategy: native
   port: 8080
+  mime-mappings:
```
wariuccio, line 23: Re-create the mime mapping we had in the Nginx config file. I'm not sure if we also need to add more of them

example-docker/docker-compose.yml
```diff
@@ -11,15 +11,15 @@ services:
       - TZ=Etc/UTC
       - DATABASE_URL=jdbc:mariadb://mariadb:3306/booklore   # Only modify this if you're familiar with JDBC and your database setup
       - DATABASE_USERNAME=booklore                          # Must match MYSQL_USER defined in the mariadb container
-      - DATABASE_PASSWORD=your_secure_password              # Use a strong password; must match MYSQL_PASSWORD defined in the mariadb container 
-      - BOOKLORE_PORT=6060                                  # Port BookLore listens on inside the container; must match container port below
+      - DATABASE_PASSWORD=your_secure_password              # Use a strong password; must match MYSQL_PASSWORD defined in the mariadb container
+      - BOOKLORE_PORT=8080                                  # Port BookLore listens on inside the container; must match container port below
```
wariuccio, line 15: I am not using this variable. Do we need to? If yes we should use it as the port used by the Spring Boot application

acx10:
Thanks for the tips and comments!

I’ve created another PR with the additional changes that are necessary: https://github.com/booklore-app/booklore/pull/1766

wariuccio:
This is being done under https://github.com/booklore-app/booklore/pull/1766

PR #1682, opened 2025-11-29
Add actual language to Kobo metadata

### LINKED ISSUE #1681: Kobo Sync: Language of eBooks is incorrectly always set to "English"
# 🐛 Bug Report for Booklore

---

## 📝 What happened?

The Language of eBooks transferred via Kobo Sync is always set to "English". As a result, the **wrong dictionary** is used and the **hyphenation** is incorrect.

## 🔄 Steps to Reproduce

1. Assign a non-English eBook to the Kobo Shelf, for example German. (The language in the metadat is correctly set.)
2. Sync with the Kobo Device and let's transfer it as Kepub.
3. Open the file on the Kobo device
4. Go to the details view. You can see that the Language is set to English instead of German.
5. Long tap on a word in order to translate it. It tries to translate from English instead of German.

**Result:** 
The Book-Language is assumed to be English instead of German.

## ✅ Expected Behavior

The Language should be set correctly; in the above example to German.

## 💻 Environment

- **Booklore Version:** 1.12.0
- **OS:** Debian 13
- **Browser:** Brave
- **Installation:** Podman
- **Storage Type:** SSD

beedaddy:
Fixes #1681

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/KoboEntitlementService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/KoboEntitlementService.java
@@ -205,6 +205,7 @@ public class KoboEntitlementService {
                 .description(metadata.getDescription())
                 .contributors(authors)
                 .series(series)
+                .language(metadata.getLanguage())
                 .downloadUrls(List.of(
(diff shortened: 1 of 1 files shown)


beedaddy:
I'm not sure if you need to set the language in other places as well, but at least it fixes the mentioned issue for me.

PR #1691, opened 2025-11-30
feat(opds): Sort OPDS Feed by addedOn Desc 

WorldTeacher:
This PR changes the default sort of the OPDS Feed to sort the files in descending order, based on the addedON date. 

With this change, the newest titles will always be shown first and features like OPDS sync for Kobo / KoReader should work as expected 

This is something multiple uses have shown interest in, such as #1310 and #1468


WorldTeacher:
:tada:

Issue #1575, opened 2025-11-18
Allow merging books of different formats


astappiev:
**What problem or limitation are you encountering?**  
There is no ability to merge two different books in my library, e.g., one in `.pdf` and the other one in `.epub`

**What solution or improvement do you propose?**  
- Something like a `Merge to` button where we can search another book, maybe books with a default title or author should be first in the list.
- A cron job that merges books with the same `ISBN` but different formats.

**Additional details**  
I raise a new issue, as the previous one was marked as complete https://github.com/booklore-app/booklore/issues/221

tbleiker:
Duplicate of #489?

Edit: rephrase and correct issue nr.

piffio:
Hello,

Just to let everyone know that I've started working on a branch to prepare a PR implementing this behaviour.
This feature is a deal breaker for me to start using booklore as my main ebook management software.

I'll keep you posted as soon as I have a first draft PR that people could test.

Issue #489, opened 2025-06-12
[Feature Request] summarizing file formats


defcon1702:
It would be great if a book is collected in multiple file formats that it would show up only once and offer the different file formats as dropdown and in the details section. :)

P.S. Thanks for this awesome piece of codeart! It's a oure joy to use it!

adityachandelgit:
Thanks for the suggestion! This is a great idea and would definitely improve usability. However, it’s a bit tricky to implement reliably since we currently have no guaranteed way to know if multiple files belong to the same book, unless they have the exact same filename with different extensions.

If you have any thoughts on how we could reliably match such files, feel free to share!

tbleiker:
See here as well: #221.

tbleiker:
What do you think of the following approach?
- Only keep one type of library 
- Treat files (.pdf, .epub, .cbz, ...) with the same file name (within the same folder?)  as being the same book
- Consider the additional files for a book to be in a folder with the same name as the book's file name - placed in the same folder as the book's file(s)
- Allow multiple books (of different types) to be merged
   - Allow to choose which book the metadata should be taken from
   - As a bonus: allow to adjust the metadata before applying it
   - Give all the files the same file name as the book from which the metadata was taken
- In the "Book Details" view, allow to upload a file (of a different file type) as an "alternative format" 
   - Give this file the same name as the file name of the book it is attached to
- In the "Book Details" view, allow to upload files as "additional files"
   - Put the files in a folder with the same name as the book's file name (within the same folder)
   - Keep the file names
   - As a bonus: allow to rename the files in Booklore
- In the "Book Details" view
   - Show all available formats under "File Type"
   - For "Read" and "Quick Send" allow to chose the file type
   - For "Download" allow to chose from the available file types and the additional files (and maybe "all"?)
- When changes are made to the metadata:
   - Write the metadata to the book file(s) (if enabled in the settings)
   - Rename (and move, if necessary) the book file(s) and the folder with the additional files (if enabled in the settings)

Drawbacks of this approach
- Two rules are enforced for all users (although minimal and completely acceptable IMHO 😝):
   - **Different book formats must have the same file name**
   - **Any additional files relating to a book must be stored in a folder with the same name as the book's file(s)**
- If you want to put each book into its own folder (as done in calibre), you will end up with an additional folder for the extra files
- A book cannot have multiple "alternative formats" of the same file format - the only reason I can think of for wanting this is to have high-resolution and low-resolution versions of the book

[...]

Issue #221, opened 2025-04-26
Feature request: ability to merge books of different formats


astappiev:
I have PDF and ePub of many books that I bought, I want to keep both of them. This will be relatively easy to detect as they share the same ISBN.
It would be nice to keep shared metadata across multiple files.

More advanced use case: some books also can have additional files, e.g. O'Reilly books often also have a source code archive. I don't need a viewer for it, just would be nice to have it combined and organised with a simple ability to download.

adityachandelgit:
@werwolfby Yes, please go aheadm it’s a highly requested feature and would be a great addition. I haven’t had the time to look into it myself, so your help is much appreciated!

werwolfby:
Have some good results so far:

<img width="1162" height="858" alt="Image" src="https://github.com/user-attachments/assets/c701f2b7-1f70-4105-8e9e-cd91f6ee220a" />

adityachandelgit:
> Have some good results so far:
> 
[...]

PR #1694, opened 2025-11-30
Fix/1608 avoid fatal errors when importing multiple books

piffio:
As reported in #1608 , importing multiple files via the bookdrop feature is brittle and often leads to errors.

Some of them are unrecoverable, as they leave the DB in an inconsistent state.

This PR makes the whole process more robust by including the following changes:

1. Make AmazonBookParser more robust by filtering out occasional "null" book IDs

2. Don’t fail imports on metadata-refresh errors as these left the DB in an inconsistent state, causing "duplicate key violations" when trying to import the same file a second time. A unit test has been updated to reflect the new behaviour.

3. Make multi-file processing more robust by avoiding race conditions with the watcher

This is my first PR on this project, so any comments on how to make it better is more than welcome.

FYI, I worked on this PR as a prerequirement to then improve the multi-format support as described in #221 , #489 and #1575

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/bookdrop/BookDropService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/bookdrop/BookDropService.java
@@ -274,7 +274,8 @@ public class BookDropService {
 
     private void unregisterAffectedLibraries(Set<Long> libraryIds) {
         for (Long libraryId : libraryIds) {
-            monitoringRegistrationService.unregisterLibrary(libraryId);
+            monitoringRegistrationService.pauseLibraryMonitoring(libraryId);
+            monitoringRegistrationService.unregisterLibraryAndWait(libraryId);
             log.debug("Unregistered library {} from monitoring", libraryId);
         }
         log.info("Unregistered {} libraries from monitoring", libraryIds.size());
@@ -293,6 +294,9 @@ public class BookDropService {
                 }
             } catch (Exception e) {
                 log.warn("Failed to re-register library {} for monitoring: {}", libraryId, e.getMessage());
+            } finally {
+                monitoringRegistrationService.resumeLibraryMonitoring(libraryId);
+                log.debug("Resumed monitoring for library {}", libraryId);
             }
         }
         log.info("Re-registered {} libraries for monitoring", libraryIds.size());
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/monitoring/MonitoringRegistrationService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/monitoring/MonitoringRegistrationService.java
@@ -7,6 +7,7 @@ import org.springframework.stereotype.Service;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.Collection;
 import java.util.Map;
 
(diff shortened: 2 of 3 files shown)


balazs-szucs:
Sorry to butt-in:

Since the parsers and file management in Bookdrop have historically been fragile and bug-prone, adding JUnit tests would greatly improve stability and maintainability.

[nitpick] Also, please consider refactoring comparisons to use idiomatic modern Java methods instead of `==`, which compares references for objects and can lead to subtle bugs (e.g., `new String("a") == new String("a")` returns `false`). Better alternative Object.equals imho. But haven't ran locally so not sure how valid this advice is.

> This PR makes the whole process more robust by including the following changes:
> 1. Make AmazonBookParser more robust by filtering out occasional "null" book IDs
> 2. Don’t fail imports on metadata-refresh errors as these left the DB in an inconsistent state, causing "duplicate key violations" when trying to import the same file a second time. A unit test has been updated to reflect the new behaviour.
> 3. Make multi-file processing more robust by avoiding race conditions with the watcher

As the saying goes, "Pics or it didn't happen", but in this case, "JUnit tests or it didn't happen.". Personally I feel like your parser changes might not exactly do what you think they do, but I can be easily proven wrong :)

FYI, I am just contributor, I just wrote too many JUnit tests in past few releases, I would like to see others follow suit. Especially, if you intend to make follow-up PRs as you alluded in your PR message.

piffio:
Thanks @balazs-szucs for taking the time to write your comment.

TBH, this is exactly the kind of feedback I was looking for in this early PR, as a way to better understand the project expectations and unwritten standards.

A few follow-up comments

**Bookdrop fragility** 
I see what you're talking about, and that's the main reason why I wanted to contribute such fixes. I do believe that he bookdrop experience must be smooth and flawless, as that's often the first thing users will see. I got annoyed by the failures I noticed, and I decided to dig into the code to solve them.
Indeed, it makes a lot of sense to add JUnit tests for such fixes. I am currently unsure about whether I'd be able to fully reproduce the underlying issue in unit tests, but I'll definitely try and report back.

**Idiomatic Java** 
It's been more than a decade since I last did any meaningful Java work, so these comments are very helpful as they're giving a clear direction on expectations. Point well taken

**Parser logic**
Not here to prove anyone wrong, and I might very much be wrong -- or imprecise -- myself.
Again, I'll do a better job at reproducing the issue first and see if I can come up with a way to capture it in a JUnit test with the same or a more appropriate fix

Expect a rework in the upcoming days, unless life gets in the way!

balazs-szucs:
No worries at all!

I just wanted to clarify that I intended my previous feedback to be a bit lighthearted. Re-reading it, I realized the cheeky tone might not have come across as intended. Just to be clear: I do believe you are on the right track :)

My main concern was simply that `==` compares references in Java (AFAIK), which might not be sufficient here. However, I haven't pulled down your PR to test it locally. I just made that observation by looking at the changes, so I could be wrong (e.g., the difference between the String "null" vs an actual null object).

That is why JUnit tests would be great. They would definitely establish:
- Whether this is an edge case that needs to be considered.
- If the proposed fix actually resolves the issue

piffio:
@balazs-szucs Ignore this PR for now. I'll close it soon.
Have a look at #1709 instead, which should contain a proper fix.

You were right: though the original fix in this branch worked, it did so by masking the underlying problem, which seems to be fixed now.

piffio:
Superseded by #1709

PR #1696, opened 2025-11-30
Downgrade Java, Spring Boot and Gradle

acx10:


Muppetteer:
This commit has broken the dev docker environment. It can't compile presumably because it uses a gradle:9-jdk25-alpine image

PR #1728, opened 2025-12-02
Add filter mode to querystring

Muppetteer:
Fixes #1719 

1. Adds filter mode to the query string so it can be preserved on page reload or link sharing.
2. Changes the filter accordion to multiple mode so all selected filters are expanded and visible.
3. Skips a bunch of unnecessarily repeated processes by checking if filter mode is actually changed when handling onchange events.

### DIFF
--- a/booklore-ui/src/app/features/book/components/book-browser/book-filter/book-filter.component.scss
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-filter/book-filter.component.scss
@@ -36,7 +36,6 @@
   flex: 1;
   overflow-y: auto;
   padding-bottom: 1rem;
-  padding-right: 0.25rem;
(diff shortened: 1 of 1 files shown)


acx10:
@Muppetteer I noticed an issue that I believe started when the “Single” filter was introduced. When you’re on the /all-books page and refresh the browser, the filter icon becomes active automatically, which shouldn’t happen. Here’s a video demonstrating the issue:


https://github.com/user-attachments/assets/bac54270-16da-4d41-9a1c-b99673354a34

Muppetteer:
Ok I'll look into it

Muppetteer:
It should be fixed now. There was a bit of a nasty event ordering issue that I had to untangle.

Issue #1724, opened 2025-12-02
Make it possible to somehow mark empty metadata fields as "correct"


RonnyAL:
# ✨ Feature Request for Booklore

Thank you for contributing to Booklore's development. Your suggestions help shape the future of this project.

---

## 📝 Description
While looking into how to improve metadata quality, I found no way to tell the system that "this title definitely has no subtitle" - so the lack of one detracts from the calculated metadata quality. I could of course set the weighting of the subtitle field to 0, but that would artificially inflate the matadata quality calculation for books that are *actually* missing data.

## 🌟 Benefits
This would make it possible, at least through some manual work, to elevate the metadata quality rating, not artificially (through lowering weightings), but in a real way that reflects verified data.

## 💻 Use Cases
Many titles refrain from specifying certain fields. I believe the subtitle is a common one, although there are probably other examples I can't think of right now. I suppose there would ideally be a process determining which fields should have this option.

## 📸 Visual References _(Optional)_
-

## 🎨 Technical Details _(Optional)_
My best idea (something I also tried), is that one could indicate an empty value as correct by locking the (empty) field. That would require no UI changes.

## 📌 Additional Context
-

PR #1729, opened 2025-12-02
refactor(metadata): consider locked, but empty or null metadata valid

balazs-szucs:
This pull request enhances the logic for calculating metadata match scores in `MetadataMatchService` by factoring in "locked" fields, and adds new unit tests to verify this behavior. The main change is that fields marked as "locked" are now considered present for scoring purposes, even if their value is empty or null. This ensures that locked fields contribute to the match score, improving accuracy in scenarios where certain metadata fields are intentionally fixed.

**Metadata match score logic improvements:**

* Updated `calculateMatchScore` in `MetadataMatchService.java` to treat fields as present if they are marked as "locked", by modifying helper methods (`isPresent`, `hasContent`, `isPositive`) to accept a `locked` parameter and return true if locked.

**Testing enhancements:**

* Added `MetadataMatchServiceTest.java` to verify that locked fields are correctly scored, including tests for empty locked subtitle and null locked series number.

Closes: #1724

### DIFF
--- a/booklore-ui/src/app/features/book/components/book-browser/book-card/book-card.component.scss
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-card/book-card.component.scss
@@ -234,3 +234,7 @@
 ::ng-deep .custom-button-padding .p-button {
   padding-block: 0.25rem !important;
 }
+
+::ng-deep .book-details-dialog .p-dialog-header {
+  padding: 1rem 1.25rem !important;
+}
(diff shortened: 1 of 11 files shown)


acx10:
@balazs-szucs Nice thinking! So basically, if a user locks a field, we treat it as if that field has a value for scoring purposes.

RonnyAL:
Thanks for the very quick implementation of this feature request. I'm new to Booklore, and so happy to see how responsive the maintainers are.

balazs-szucs:
Glad I could help. 

If the issue still persist (after release, it's only in the develop branch right now), or you have another issue feel free to ping or make another ticket.

PR #1739, opened 2025-12-03
Feature: filter mode preference

### LINKED ISSUE #1720: [Feature] Add default filter type to settings
# ✨ Feature Request for Booklore

Thank you for contributing to Booklore's development. Your suggestions help shape the future of this project.

---

## 📝 Description

As requested in #1640 and implemented through #1671 there's now a single select filter type, which is great.
As this is my default style of browsing, I would like it to be a setting.

## 🌟 Benefits

Setting single select as default filter type prevents switching every time

## 💻 Use Cases

1. For people who like the single select to be the default behaviour
2. 
3.

Muppetteer:
Fixes #1720 

While doing this I noticed there is also a preference for FILTER_SORTING_MODE (alphabetical or count) being written to the database but it is never exposed anywhere in the UI. It should not be too difficult to add to this if wanted.

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/controller/MetadataController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/controller/MetadataController.java
@@ -72,8 +72,8 @@ public class MetadataController {
                 .updateThumbnail(true)
                 .mergeCategories(mergeCategories)
                 .replaceMode(MetadataReplaceMode.REPLACE_ALL)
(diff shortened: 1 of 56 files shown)


acx10:
@Muppetteer Actually I had added FILTER_SORTING_MODE in the past but removed it because it didn’t feel very beneficial at the time. But yeah, I think it makes sense to bring it back, and defaulting to alphabetical sounds good. Thanks for pointing it out!

Issue #1730, opened 2025-12-02
[Feature] Make search diacritic insensitive


lbird-mma:
# ✨ Feature Request for Booklore

Thank you for contributing to Booklore's development. Your suggestions help shape the future of this project.

---

## 📝 Description

When searching for an author (or anything else for that matter) which has a diacritic in it's name, the search without diacritic returns nothing. For example I have books from author 'Jo Nesbø', but searching for 'nesbo' returns nothing. I like to have correct names for authors (etc), but when searching it's sometimes hard to type.
Search could be diacritic- or accent-insensitive.

## 🌟 Benefits

Search is easier

## 💻 Use Cases

1. 
2. 
3.

PR #1791, opened 2025-12-08
feat(search): Add denormalized search_text column for better book search

balazs-szucs:
Adds a pre-computed `search_text` column to the `book_metadata` table to optimize book search queries. Search text is normalized at write time, eliminating expensive JOINs and `LOWER()` operations at query time.

### Backend
- New column: `search_text` (TEXT) in `book_metadata` table
- Auto-update: `@PrePersist`/`@PreUpdate` hooks rebuild search text on metadata changes
- Manual triggers: Added `updateSearchText()` calls where author collections are modified (since JPA lifecycle hooks don't fire on collection-only changes)
- Normalization: Unicode diacritics (é→e, ø→o, ß→ss, ł→l, æ→ae, œ→oe) are normalized for accent-insensitive search
- Programming languages: Preserved `#` and `+` characters for book titles like "C++", "C#", "F#"

### Frontend
- Updated `HeaderFilter` normalization to match backend logic

### Queries Simplified
```diff
- LEFT JOIN m.authors a
- WHERE LOWER(m.title) LIKE LOWER(CONCAT('%', :text, '%'))
-    OR LOWER(m.subtitle) LIKE LOWER(CONCAT('%', :text, '%'))
-    OR LOWER(m.seriesName) LIKE LOWER(CONCAT('%', :text, '%'))
-    OR LOWER(a.name) LIKE LOWER(CONCAT('%', :text, '%'))
+ WHERE m.searchText LIKE CONCAT('%', :text, '%')
```

Closes: #1730

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/repository/BookRepository.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/repository/BookRepository.java
@@ -74,9 +74,9 @@ public interface BookRepository extends JpaRepository<BookEntity, Long>, JpaSpec
 
     @Query(value = """
                 SELECT DISTINCT b.* FROM book b
-                LEFT JOIN book_metadata m ON b.book_id = m.book_id
+                LEFT JOIN book_metadata m ON b.id = m.book_id
                 WHERE (b.deleted IS NULL OR b.deleted = false)
-                ORDER BY b.book_id
+                ORDER BY b.id
                 LIMIT :limit OFFSET :offset
             """, nativeQuery = true)
     List<BookEntity> findBooksForMigrationBatch(@Param("offset") int offset, @Param("limit") int limit);


acx10:
@balazs-szucs I’m not sure if adding a dedicated search column is the best approach for a couple of reasons:
	1.	The search text would need to be recalculated on every metadata update. While the query itself isn’t heavy, it still adds overhead.
	2.	We’d need to ensure the recalculation method is always invoked correctly, which could be error-prone.

Wouldn’t it be cleaner to handle diacritics normalization directly within the existing SQL queries instead?

acx10:
@balazs-szucs I am open for discussion though.

balazs-szucs:
From an end-user perspective, users typically prefer to trade some initial setup time or write-time overhead for faster runtime performance. The pre-computed search text approach is better for this; it provides instant search results that make the application feel better on an _every-day_ basis, while metadata updates just run in the background where nobody generally cares about the overhead. (At least I think so :D, regarding my personal instance, there are quite a few people who use it who would benefit from "faster" search, but generally wouldn't know about the operational side, e.g., scans, etc.). But I do had to think about this also tbh.

balazs-szucs:
As for the overhead on keeping the data and DB consistent also 100% valid, but that's our burden that we _can_ manage for improved user experience, I think.

balazs-szucs:
Although I do wonder about actual real-world numbers. I would love see how this pans out on BL instance with a large library (meaning with actual real numbers), but my intuition says, it this is the "better" trade off, at least for most people.

acx10:
Alright, you’ve convinced me! :) 
Your reasoning makes sense. I’ll pull the PR locally, run some additional manual tests, and proceed with merging it if everything checks out. Thanks for the thorough explanation and great work!

balazs-szucs:
Apologies for the messy git handling, somewhere along the line there was `V66__Add_icon_type_to_entities` and I added originally, `V67__Add_search_text_column` as V66, so I had to quickly update it. Should be fixed now. AFAICT this should be working now :)

lbird-mma:
Not sure about all your considerations, but as a SQL Server developer I was thinking more of using accent- and case insensitive collations in your WHERE clauses, like this

`SELECT * FROM my_table WHERE name COLLATE utf8mb4_0900_ai_ci = 'search_term';`

This way you can still query against your original data

PR #1799, opened 2025-12-09
Fix: locking fields discards unsaved changes

### LINKED ISSUE #1798: [BUG] locking fields discards unsaved changes
## 📝 What happened?
While working on the metadata edit form I noticed than any form of lock toggle discards unsaved changes on the form. Lock All potentially nukes a whole page worth of changes without any warning.

## 🔄 Steps to Reproduce
1. Open metadata edit form for a book 
2. Change the value of a field
3. Press Lock on the field - or press Lock All at bottom

**Result:**
Field(s) are locked and changes are lost.

## ✅ Expected Behavior
Changes should be saved and fields locked.

## 💻 Environment
- **Booklore Version:** v1.13.2

## 📌 Additional Context
A little debugging shows that the form is submitted with correct data to the BookService. I believe there is probably an ordering bug in there somewhere and it thinks that a field or all fields are locked and discards the changes.

Possibly related to changes made in #1672 ?

## ✨ Possible Solution _(Optional)_
A dirty hack I've played with is to save the form first then save it again with fields locked, but it would be better to solve this properly.

Muppetteer:
Fixes #1798


balazs-szucs (commented):
Nice find :)

Can you add a JUnit test to cover this edge case?

PR #1802, opened 2025-12-09
 refactor(api): minor QOL updates, code modernization 

balazs-szucs:
This pull request introduces several quality-of-life improvements, entity stability features, and minor code modernizations across the codebase.

The key change is ensuring data integrity for new persistence entities by explicitly setting a default `iconType` and improving RAR file metadata handling.

#### Entity Persistence and Data Integrity
* **Default IconType:** Added `@Builder.Default` and `@PrePersist` logic to `LibraryEntity`, `MagicShelfEntity`, and `ShelfEntity` to ensure the `iconType` field is defaulted to `IconType.PRIME_NG` if it is null upon saving.
* **Equals Method Fixes:** Updated `equals()` implementations in several entities (`AuthorEntity`, `CategoryEntity`, `MoodEntity`, `TagEntity`) to consistently reference the `id` field instead of the getter, improving clarity and consistency.

#### Code Modernization and Cleanup
* Replaced index-based list access (`.get(0)`) with the more expressive `Collection.getFirst()` in relevant services and tests (requires Java 21 or later).
* Simplified boolean return statements in `CbxConversionService` and `KoboReadingStateService`.
* Broadened type acceptance in `BookRuleEvaluatorService.toStringList` from `List` to `Collection`.
* Removed unnecessary checked exceptions (`throws IOException`, `throws Exception`) from controller methods, indicating better internal exception handling.

#### CBX Metadata Writer Refactor
* Simplified the `CbxMetadataWriter` logic for handling **RAR file** extraction and re-archiving, primarily by removing a redundant inner `try-finally` block for cleaner code flow.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -190,7 +190,7 @@ Please be respectful, inclusive, and collaborative. Harassment, abuse, or discri
 
 ## 📄 License
 
-Booklore is open-source and licensed under the MIT License. See [`LICENSE`](./LICENSE) for details.
+Booklore is open-source and licensed under the GPL-3.0 License. See [`LICENSE`](./LICENSE) for details.
 
 ---
 
--- a/booklore-ui/src/app/features/book/components/book-browser/book-browser.component.html
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-browser.component.html
@@ -4,21 +4,28 @@
     <div class="flex items-center justify-between w-full gap-4 py-2 px-2 md:px-4 rounded-t-xl md:mb-4 mb-2 bg-[var(--card-background)] border-[1px] border-solid border-[var(--p-content-border-color)]">
       <div class="flex items-center min-w-0 max-w-full flex-shrink">
         @if (entityType$ | async; as entityType) {
-          <p class="text-xl whitespace-nowrap overflow-hidden text-ellipsis w-full">
-            @if (entityType === EntityType.ALL_BOOKS) {
-              @if (currentFilterLabel && rawFilterParamFromUrl) {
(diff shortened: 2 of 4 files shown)


booklore-api/src/main/java/com/adityachandel/booklore/model/entity/ShelfEntity.java
```diff
@@ -47,4 +48,11 @@ public class ShelfEntity {
     )
     @Builder.Default
     private Set<BookEntity> bookEntities = new HashSet<>();
+
+    @PrePersist
+    public void ensureIconType() {
```
acx10: Is @PrePersist really needed as we're explicitly setting the icon type 'private IconType iconType = IconType.PRIME_NG;'

balazs-szucs: Probably not, I guess only way it would be relevant if some would set it explicitly to null, but then I guess there is intent there. I'll update it. Thanks for the review :)

balazs-szucs: Addressed in: [d5f7512](https://github.com/booklore-app/booklore/pull/1802/commits/d5f7512688132ff4352401158c2a26eafe4e098c)

acx10: > Probably not, I guess only way it would be relevant if some would set it explicitly to null, but then I guess there is intent there. I'll update it. Thanks for the review :)

NOT NULL constraints in the table should take care of it though the error message won't be pretty.

PR #1810, opened 2025-12-10
Fix vertical overflow on metadata edit form dialog

Muppetteer:
Fixes #1540 - the metadata edit form was vertically overflowing its container when displayed in dialog mode.

Also contains a minor fix to wrapping of tag chips and clear button inside autocomplete fields on the same form.

### DIFF
--- a/booklore-ui/src/app/features/metadata/component/book-metadata-center/book-metadata-center.component.scss
+++ b/booklore-ui/src/app/features/metadata/component/book-metadata-center/book-metadata-center.component.scss
@@ -1,9 +1,28 @@
-.tabpanels-responsive {
+::ng-deep .layout-main .tabpanels-responsive {
   height: calc(100dvh - 9.7rem);
 }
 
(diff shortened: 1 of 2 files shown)


acx10:
This PR has introduced some UI issues, as shown in the video. I believe the following CSS is the likely culprit, as it affects more components than intended:

```
::ng-deep .p-dialog-content,
::ng-deep p-dialog app-book-metadata-center,
::ng-deep p-dialog app-book-metadata-center > div {
  display: flex;
}
```

It appears to have unintended side effects on other dialogs as well, including the icon selector in some cases. To avoid further impact, I’m reverting the PR for now.


https://github.com/user-attachments/assets/0f64fa7d-98d0-4efa-b665-30f6a62e0df7

Muppetteer:
Looking into it. Unfortunately there's no way to know which dialog you have open through the CSS as they don't have an ID attribute or helpful class, so that makes it difficult to target a specific one like this.

PR #1819, opened 2025-12-11
Custom metadata provider

FlyinPancake:
This PR adds a custom metadata provider with an API spec that can be implemented by anyone to add a metadata provider without needing to integrate into BookLore

### DIFF
--- a/booklore-api/.gitignore
+++ b/booklore-api/.gitignore
@@ -35,3 +35,5 @@ out/
 
 ### VS Code ###
 .vscode/
+
+appdata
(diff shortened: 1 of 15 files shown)


FlyinPancake:
So I think there is a mismatch between our understanding this spec, so here's a breakdown:

The community would create metadata provider servers that users can host as sidecars to BookLore. 
Take a look at what the ABS community achieved in ~2 years: https://www.audiobookshelf.org/guides/custom-metadata-providers/#community-providers

I will make a reference implementation (most likely in python) so developers will have a starting point / reference point.

BookLore users would only need to enter 3 things: provider name, endpoint URL and auth token.

Originally the ABS CMP spec was created by me and a friend of mine. This spec follows that spec in spirit. This spec is larger, because BookLore supports richer metadata and the ABS one does not have any examples / descriptions that would help developers implement required changes.

CMP developers could omit the majority of fields if their provider does not support that feature (i.e.: ratings) or they simply don't want to build it.

FlyinPancake:
Not ready yet, somehow I broke loading settings...

grotell:
Any updates on this PR?

FlyinPancake:
closing in favor of #2693

Issue #1762, opened 2025-12-05
[Feature] Make personal rating personal


lbird-mma:
# ✨ Feature Request for Booklore

Thank you for contributing to Booklore's development. Your suggestions help shape the future of this project.

---

## 📝 Description

Books have a personal rating in their metadata, which is not at all personal. In a multi-user environment everyone sees the same rating and another user can reset my personal rating. Maybe it's a good idea to pull personal rating from the metadata and make it really personal

## 🌟 Benefits

Prevents overwriting someone's personal rating

## 💻 Use Cases

1. Multi-user environment can have true personal ratings.
2. 
3.

lbird-mma:
Sort of duplicate. #1492 is created as a bug (which it is not, considering it's metadata). This a feature request

Muppetteer:
Issues I can see after briefly looking at this one:

- Per user/book rating either needs its own table or a column added to `user_book_progress` which is already per user/book. The latter would be the simple solution. I guess it is kind of related to progress as you normally rate after reading.
- Migration of existing ratings? If there is only one user it would be simple but if there are multiple then there is no perfect solution. Either migrate to first user/a defined user or just ignore them.
- Sync to/from Kobo / Koreader etc (if possible)? If you do that then potentially you could have multiple different ratings for the same book/user being synced back from devices.

PR #1820, opened 2025-12-11
Per user personal ratings

Muppetteer:
As requested at #1762 

Currently "personal ratings" in Booklore are not personal at all - they are saved as book metadata and written to book files, both of which are shared amongst Booklore users. This patch aims to correct that and does the following:

- Moves the `personal_rating` column from `book_metadata` to the `user_book_progress` table.
- Copies ratings from `book_metadata` to every user that already has a progress record in `user_book_progress` for the matching book.
- Removes `personal_rating` and `personal_rating_locked` columns from `book_metadata`.
- Stops reading and writing ratings from or to book file metadata.
- Adds an option to sort by personal rating in the library UI.
- Fixes the personal rating filter so that it works in the library UI.
- Preserves personal rating functionality on the charts page, only now it is user specific.

Consequences:
- The last user that rated a book before this patch will win over any other users that also rated it, as they will all receive the same rating for the book. There is nothing I can do about that.
- Books that are rated without a matching `user_book_progress` record will not be rated in Booklore after this change. In an earlier version of this patch I had an extra query in the DB migration script that creates a `user_book_progress` record for every rated book and user combination. At the risk of spamming the table with unwanted ratings, that might be a more desirable option for anyone worried about losing those ratings (let me know if so).
- Personal ratings are no longer written to book metadata, nor are they read from books when imported. This might mean a book that has a rating written to its file metadata (e.g. 9) will not have its metadata rewritten when rated 8 in Booklore for example.

Other notes:
- I discovered that there is an unused `rating` column on the `book_metadata` table and also an unused `read_status` column on the `book` table. I can remove those as part of this update if they are no longer necessary.

As this touches a lot of files and requires database alteration, I would appreciate some feedback and testing (especially the DB migration script) before merging.

### DIFF
--- a/booklore-ui/src/app/features/book/components/book-browser/book-card/book-card.component.ts
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-card/book-card.component.ts
@@ -179,19 +179,7 @@ export class BookCardComponent implements OnInit, OnChanges, OnDestroy {
         icon: 'pi pi-info-circle',
         command: () => {
           setTimeout(() => {
-            if (this.metadataCenterViewMode === 'route') {
-              this.router.navigate(['/book', this.book.id], {
-                queryParams: {tab: 'view'}
-              });
-            } else {
-              this.dialogService.open(BookMetadataCenterComponent, {
-                width: '95%',
-                data: {bookId: this.book.id},
-                modal: true,
-                dismissableMask: true,
-                showHeader: false
-              });
-            }
+            this.openBookInfo(this.book);
           }, 150);
         },
       },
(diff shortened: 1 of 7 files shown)


beedaddy:
Thanks a lot for that. No problems with it in my small dev environment (I can't comment on the DB migration from v1.13.2.). Not copying ratings without a `user_book_progress` seems perfectly reasonable to me.

Muppetteer:
> Personal ratings are no longer written to book metadata, nor are they read from books when imported. This might mean a book that has a rating written to its file metadata (e.g. 9) will not have its metadata rewritten when rated 8 in Booklore for example.

Shower thought on this one: create a copy of the Booklore file and write the user's individual rating to it before syncing it to a Kobo or emailing to Kindle. That way each user would get an individual copy of an epub or whatever with their rating applied to it. I'm unsure how useful this would be though as you would normally rate a book after reading it.

adityachandelgit (approved):

acx10:
> > Personal ratings are no longer written to book metadata, nor are they read from books when imported. This might mean a book that has a rating written to its file metadata (e.g. 9) will not have its metadata rewritten when rated 8 in Booklore for example.
> 
> Shower thought on this one: create a copy of the Booklore file and write the user's individual rating to it before syncing it to a Kobo or emailing to Kindle. That way each user would get an individual copy of an epub or whatever with their rating applied to it. I'm unsure how useful this would be though as you would normally rate a book after reading it.

I’d avoid adding that level of complexity. A book file really shouldn’t contain user-specific metadata, especially since the same file may be used by multiple people. Keeping personal ratings separate from the actual book file is cleaner.

PR #1824, opened 2025-12-11
feat(opds): allow user to set sorting for opds feed in settings

WorldTeacher:
first off: sorry for the squash-like commit. I messed up my local branches and instead of trying to fix that, I opted to simply extract the changed files and start over with a new fork.

This PR adds support to select a sort mode for the opds feed. In this PR, these sorting options have been added:
- Recently added
- Title (A-Z / Z-A)
- Author (A-Z / Z-A)
- Series (A-Z / Z-A) -> this first sorts by the series (in series it sorts by number), then sorts all entries without series by addedOn
- Rating (asc, desc) -> rating is calculated like this: `SUM(hardcover, goodreads, amazon) / count(ratings not zero)`

The sort setting has been added into the "Create OPDS User" section
image
and uses `Recently Added` by default. 
Once the user has been added, the sort order can be changed using the pencil icon
image

After the save, it takes ~2-3s for the changes to be applied, at least in my local tests using Thorium reader.

Feel free to rename the migration sql file - I just picked that number to get out of the way of the current migration files


booklore-api/src/main/java/com/adityachandel/booklore/service/opds/OpdsBookService.java
```diff
@@ -79,19 +79,19 @@ public Page<Book> getBooksPage(Long userId, String query, Long libraryId, Long s
         if (libraryId != null) {
             validateLibraryAccess(libraryId, userLibraryIds, isAdmin);
             Page<Book> books = query != null && !query.isBlank()
-                    ? searchByMetadataInLibrariesPageInternal(BookUtils.normalizeForSearch(query), Set.of(libraryId), page, size)
```
acx10, line 82: Hi, I noticed that BookUtils.normalizeForSearch() was removed in this PR. Could you clarify the reason for its removal?

acx10:
Everything else looks great, aside from the one question I left on the PR, and merge conflicts. Thanks for your work on this!

booklore-api/src/main/java/com/adityachandel/booklore/service/opds/OpdsBookService.java
WorldTeacher: Hi, sorry for the confusion and removal. I looked at the code and, embarrassingly, this came from me messing up the local branches. I went back through the original files and even in my first commit, the query wasn't normalized. Then I missed this when adding the old files into the new fork. I have re-added the normalization in a new commit. Sorry for the trouble.

Issue #1823, opened 2025-12-11
Failed to download kobo book CBZ when clicking download in device


thvardhan:
# 🐛 Bug Report for Booklore

Thank you for taking the time to report this bug. Your feedback helps make Booklore better for everyone! 

Let's squash this bug together! 🔨

---

## 📝 What happened?
Not able to download books to kobo with kobo sync. The books show up on device but clicking download fails.

## 🔄 Steps to Reproduce
I tested booklore a few hours back with a CBZ file with kobo sync. everything was fine, my single file got converted to epub automatically and was synced to my device. all good. I moved in my entire library having around 50 cbz files. all are in kobo shelf. now when I click download on any book, the device shows downloading and then it says sync completed but the book is not actually downloaded. I checked the docker logs and I see a bunch of errors which is attached as logfile. strange thing is, now I just removed all books kept 1 in my kobo shelf yet I still get this error.

**Result:** file is not downloaded.

## ✅ Expected Behavior
File should have been downloaded

## 📸 Screenshots / Error Messages
docker logs
[kobofail-log.txt](https://github.com/user-attachments/files/24107928/kobofail-log.txt)

## 💻 Environment
- **Booklore Version:** 1.32.2
- **OS:** Ubuntu 24.04.2 LTS
- **Browser:** Chrome on PC, kobo otherwise.
- **Installation:** Docker
- **Storage Type:** Local HDD

## 📌 Additional Context
I have noticed that when the device shows downloading, if I see the logs, it shows starting cbx conversion to epub. but before it can finish and serve, the kobo device changes its state from "downloading" to again "click to download". I think its not waiting long enough from device.

Edit : Can confirm this is the issue. I have uploaded a small cbz file and its downloaded perfectly fine to the device.

## ✨ Possible Solution _(Optional)_
Maybe store the converted epub somewhere and serve it instead of converting each time whenever download is pressed

PR #1838, opened 2025-12-12
Persist conversions of CBX to epub for future use and kobo sync

thvardhan:
This is a PR to add a fix for #1823 . I have tested this on my own kobo device and found it to be working. 
I've added a new setting called persist conversions
image
when it is turned on, a new shelf is created
image
now whenever kobo requests a cbx and it goes to conversion, after conversion is done, it would save the new file in the same library, with shelf as conversion with same metadata as original file. this allows the backend to serve the .cbx.epub files directly next time without conversion which was the reason of timeout on my device. 

future plan is to have an auto conversion button so any book added to kobo gets automatically converted before making the request to get the book.

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/book/BookDownloadService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/book/BookDownloadService.java
@@ -111,6 +111,7 @@ public class BookDownloadService {
         boolean convertCbxToEpub = isCbx && koboSettings.isConvertCbxToEpub() && bookEntity.getFileSizeKb() <= (long) koboSettings.getConversionLimitInMbForCbx() * 1024;
         boolean persistConversion = koboSettings.isPersistConversion();
 
+        int compressionPercentage = koboSettings.getConversionImageCompressionPercentage();
         Path tempDir = null;
         try {
             File inputFile = new File(FileUtils.getBookFullPath(bookEntity));
@@ -159,7 +160,7 @@ public class BookDownloadService {
             }
 
             if (convertCbxToEpub && !convertedExists) {
-                fileToSend = cbxConversionService.convertCbxToEpub(inputFile, tempDir.toFile(), bookEntity);
+                fileToSend = cbxConversionService.convertCbxToEpub(inputFile, tempDir.toFile(), bookEntity,compressionPercentage);
             }
 
             if (convertEpubToKepub && !convertedExists) {
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,37 +2,31 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: ''
+labels: bug
 assignees: ''
-
 ---
 
-# 🐛 Bug Report for Booklore
+# 🐛 Bug Report
 
-Thank you for taking the time to report this bug. Your feedback helps make Booklore better for everyone! 
-
-Let's squash this bug together! 🔨
-
----
-
-## 📝 What happened?
+## 📝 Description
 <!-- Describe what went wrong -->
 
 
 ## 🔄 Steps to Reproduce
 <!-- Provide detailed steps to reproduce the behavior. Be specific about what you clicked, typed, or configured -->
-1. 
-2. 
-3. 
-4. 
+1.
+2.
+3.
+4.
 
 **Result:** <!-- What happened after these steps? -->
+
 
 ## ✅ Expected Behavior
 <!-- Describe what should have happened instead -->
 
 
-## 📸 Screenshots / Error Messages
+## 📸 Screenshots / Error Messages _(Optional)_
 <!-- Share any screenshots or error messages here (just drag & drop) -->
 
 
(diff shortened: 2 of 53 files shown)


thvardhan:
yes I agree with you, as a matter of fact I was already making some test cases. do note that I do not have too much experience writing test cases so a review goes a long way! but here is the gist of test cases that I've added

`BookDownloadServiceKoboTest.java`

- downloadKoboBook_persistConversion_true_callsImportAndUsesConversionShelf - CBX to EPUB with persistConversion=true moves the converted file into the library and calls import using the conversion shelf.
- downloadKoboBook_persistConversion_false_streamsWithoutImport - CBX to EPUB with persistConversion=false streams the converted file without persisting or importing it.
- downloadKoboBook_existingConvertedFile_usesExistingAndNoImport - If a converted EPUB already exists in the library, it is used and conversion/import are skipped.
- downloadKoboBook_importFails_streamsAnyway - If import fails after conversion, streaming still occurs and the converted file is moved/served (failure is logged).
- downloadKoboBook_epubToKepub_persistConversion_true_importsKepub - EPUB to kepub with persistConversion=true converts, moves the kepub into the library, and imports it.

`BookImportServiceTest.java`

- importFileToLibrary_happyPath_importsAndAppliesMetadataAndAddsToShelf - Verifies the happy path: processor produces a Book DTO, metadata is applied, the book is added to the requested shelf, and the DTO is returned.
- importFileToLibrary_libraryNotFound_throwsApiError - Ensures importing into a non-existent library throws an API exception
- importFileToLibrary_invalidFileExtension_throwsApiError - Ensures files with unsupported extensions (e.g., .abc) cause an API exception for invalid file format.
- importFileToLibrary_missingBookEntityAfterProcessing_throwsApiError - Verifies that if processing returns a Book DTO but the corresponding BookEntity is not found, an API error is thrown.
- importFileToLibrary_processorThrows_propagatesException - Ensures exceptions thrown by the file processor propagate out of the import call.
- importFileToLibrary_metadataNullAndNoShelf_doesNotApplyMetadataOrShelf - Verifies that when no metadata and no shelfId are provided, no metadata update or shelf assignment occurs and import still returns the DTO.
- importFileToLibrary_shelfIdProvided_butShelfNotFound_noSave - Ensures that if a shelfId is provided but the shelf cannot be found, the book is not assigned/saved to any shelf and import returns successfully.
- importFileToLibrary_shelfSaveThrows_logsAndImportContinues - Simulates bookRepository.save(...) throwing while assigning a shelf; verifies the save is attempted (and may throw) but the import still returns the Book DTO (failure is logged/handled).

`ConversionShelfEventListenerTest.java`

- whenPersistConversionTrue_andShelfMissing_shouldCreateShelf - When persistConversion=true and conversion shelf is missing, the listener requests creation of the conversion shelf.
- whenPersistConversionFalse_andShelfExists_shouldDeleteShelf - When persistConversion=false and conversion shelf exists, the listener deletes it.
- whenPersistConversionMissing_shouldNotCrashOrAct - When persistConversion is absent, the listener safely does nothing (no crash, no create/delete).

booklore-api/src/main/java/com/adityachandel/booklore/service/book/BookDownloadService.java
```diff
@@ -95,25 +109,109 @@ public void downloadKoboBook(Long bookId, HttpServletResponse response) {
...
                 fileToSend = cbxConversionService.convertCbxToEpub(inputFile, tempDir.toFile(), bookEntity);
             }
 
-            if (convertEpubToKepub) {
+            if (convertEpubToKepub && !convertedExists) {
                 fileToSend = kepubConversionService.convertEpubToKepub(inputFile, tempDir.toFile(),
                     koboSettings.isForceEnableHyphenation());
             }
 
+            if (!fileToSend.equals(inputFile) && !convertedExists && persistConversion) {
```
acx10, line 171: From the code, it looks like the converted EPUB/KEPUB would coexist alongside the original CBX/EPUB file. That doesn’t seem ideal, since the UI would then treat them as two separate book entities.

This could introduce confusion and duplication unless there’s a clear strategy for how these files are managed and presented.

thvardhan: yes that is indeed the case, the idea was to keep a copy and put it in a shelf so it can be downloaded by user if required or delete if they want it to. This would only happen if the persist conversion is checked. another method of resolving the presentation would be to store it somewhere else and showing that the original book has an available conversion which can then be downloaded or deleted. for me, the latter is less intuitive as I would find that some random books have conversions and some don't.

MaximUltimatum: > From the code, it looks like the converted EPUB/KEPUB would coexist alongside the original CBX/EPUB file. That doesn’t seem ideal, since the UI would then treat them as two separate book entities.
> 
> This could introduce confusion and duplication unless there’s a clear strategy for how these files are managed and presented.

Can we store conversions in a "conversions" directory somewhere that is deliberately not scanned? We could put it at root level under `/books`
Then when a request comes in, just check the conversions directory to see if it was already converted. If not, start conversion.
First request will time out, 2nd will work fine.

MaximUltimatum: Also, couldn't we pre-empt conversation by doing it when they place the book on the Kobo shelf (since that indicates a _strong_ signal that they intend to download it)

acx10:
Closing this PR, sorry this is not in the scope of booklore.

PR #1842, opened 2025-12-13
fix: Consistent dialogs

### LINKED ISSUE #1763: [Feature] Make dialogs more consistent
# ✨ Feature Request for Booklore

Thank you for contributing to Booklore's development. Your suggestions help shape the future of this project.

---

## 📝 Description

I like to see more consistency in dialogs. For example I have my book information set to dialog. This dialog can only be closed by the X button in the top right corner. This is also the case for the Bulk Metadata Editor, however that one can also be closed by using the Esc key. And then there's Multi-book Metadata Editor which can be closed by clicking outside the dialog.

My personal preference: use all three possibilities, so there's an X button, we can use Esc and also we can click outside the dialog

## 🌟 Benefits

More consistent dialogs

## 💻 Use Cases

1. When editing a single book's metadata in a dialog and clicking Save, there's a toast notification popping up over the X button, which has to be closed before I can click the X button
2. 
3.

Muppetteer:
# 🚀 Pull Request

## 📝 Description
Consolidates dialog creation code and enforces consistency via default options.
All dialogs are closable via X button in top corner, clicking outside the dialog, or pressing ESC on a keyboard.
Dialogs with missing or duplicated headers corrected.

Fixes #1763

## 🛠️ Changes Made
Moved all dialog creation code into 2 existing helper classes: DialogLauncherService (general purpose dialogs) and BookDialogHelperService (book specific dialogs).

## 🧪 Testing
- [x] Tested manually in local environment
UI only - no java API changes or tests required.

Old bugged dialog:
old

New fixed dialog:
new

## ✅ Pre-submission Checklist
- [x] Branch is synced with latest `develop` branch

### DIFF
--- a/booklore-ui/src/app/features/metadata/component/book-metadata-center/metadata-editor/metadata-editor.component.scss
+++ b/booklore-ui/src/app/features/metadata/component/book-metadata-center/metadata-editor/metadata-editor.component.scss
@@ -10,6 +10,10 @@
   width: 250px;
 }
 
+::ng-deep .cover-item p-image {
+  width: 150px;
+}
+
 ::ng-deep p-image img {
   margin: 0 auto;
 }


booklore-ui/src/app/shared/services/dialog-launcher.service.ts
```diff
@@ -13,73 +26,164 @@ export class DialogLauncherService {
...
-        ...(isMobile
-          ? {
-            width: '90vw',
-            maxWidth: '90vw',
-            minWidth: '90vw',
-          }
-          : width
-            ? {width}
-            : {}),
-      },
```
acx10, line 37: @Muppetteer

I noticed that the code handling different dialog dimensions for desktop and mobile has been removed. I originally added it to reduce the excessive blank space in mobile view, so that dialogs would feel more natural on smaller screens.

Could you share how we’re handling responsive dialog sizing now? Just want to make sure we don’t end up with the same issue on mobile devices.

Thanks!

Muppetteer: Yes this sort of thing is better handled in CSS using media queries. The MQs can handle mobile but also will live adjust to a computer browser if it is resized to a small width, while doing it in the service is only going to work for the browser size at the instant that code is run.

That said I checked each dialog using browser dev tools by switching between full size and mobile mode and they all looked ok. I've committed a couple of extra CSS rules to absolutely ensure they are all exactly the same width on mobile.

One caveat is that some forms are still effectively unusable on mobile due to their layout (e.g. book metadata copy screen, metadata refresh, organize files), which they were before this patch anyway. That should be part of a separate issue as part of wider mobile friendliness that I will create.

PR #1852, opened 2025-12-13
feat(file-move): add copy mode with images and metadata duplication

dcoffin88:
# 🚀 Pull Request

## 📝 Description

Adds a copy/move mode toggle to Organize Book Files. Copy duplicates the file, metadata, and images to the target library while leaving the original untouched. This helps when using separate libraries per user (e.g., Kobo sync) and needing to duplicate shared books without re-importing or re-tagging.

## 🛠️ Changes Made

- add copy/move toggle with tooltip on Organize Book Files
- copy mode clones file, metadata, and images into the target library
- keeping original book and assets untouched

## 🧪 Testing

- [x] Tested manually in local environment (copy and move flows)
- [ ] Added/updated automated unit/integration tests
- [ ] All existing tests pass

- Tooltip next to mode selector (only visible when Copy is selected)

## ✅ Pre-submission Checklist

- [ ] Tests added/updated to cover changes
- [ ] All tests pass locally (`./gradlew test` for backend)
- [ ] Flyway migration versioning is correct _(if database changes made)_
- [x] Branch is synced with latest `develop` branch
- [ ] Documentation PR created at [booklore-docs](https://github.com/booklore-app/booklore-docs) _(for major features)_

## 💬 Additional Notes _(optional)_

Context: I use one library per user to keep Kobo sync separate. When users share books, copying between libraries avoids manually re-adding and re-tagging metadata. I know this creates duplicates in the All Books view for Admin, but that’s acceptable for my use case. Tested and working.

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/model/dto/request/FileMoveRequest.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/model/dto/request/FileMoveRequest.java
@@ -1,7 +1,7 @@
 package com.adityachandel.booklore.model.dto.request;
 
+import com.adityachandel.booklore.model.enums.FileOperationMode;
 import lombok.Data;
-
 import java.util.List;
 import java.util.Set;
 
@@ -9,6 +9,7 @@ import java.util.Set;
 public class FileMoveRequest {
     private Set<Long> bookIds;
     private List<Move> moves;
+    private FileOperationMode mode;
 
     @Data
     public static class Move {
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/file/FileMoveHelper.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/file/FileMoveHelper.java
@@ -68,6 +68,19 @@ public class FileMoveHelper {
         executeWithRetry(() -> Files.move(source, target, StandardCopyOption.REPLACE_EXISTING));
     }
 
+    public void copyFile(Path source, Path target) throws IOException {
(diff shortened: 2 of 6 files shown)


acx10:
Agree with @balazs-szucs! 

File handling is one of the most fragile and error-prone areas of Booklore. Even small changes can have unexpected side effects, especially since the file watcher runs continuously in the background and can interfere with file moves or modifications.

Because of this, I’d strongly recommend adding robust JUnit tests around the new copy logic, and also doing some manual regression testing of any existing features that could be affected by this change.

Thanks again for the contribution. I’ll take a closer look at the PR soon.

Muppetteer:
Surely a better fix for this would be to change device sync settings to be per user account rather than duplicating book files and database entries.

dcoffin88:
I was looking into this, but from my understanding each user can have only one Kobo shelf for their library. Each user sees only their books, with independent metadata.

So if the admin wants to add a book to a user’s library that already exists in another library, copying it between libraries keeps Kobo sync fully isolated per user and avoids having to re-add and re-tag metadata.

Muppetteer:
Ah I see you mean you want the same book in multiple libraries.

> I use one library per user to keep Kobo sync separate.

If that is your only reason for having multiple libraries then it is unnecessary. Kobo sync is already separate per user. Unlike in the physical world, shelves in Booklore do not relate to libraries. A user can assign books from any library that they have access to onto a shelf (including Kobo).

Just ensure that the auto add to Kobo option is off and users can manually add whatever books they want to their Kobo shelf from your shared librar(ies).

PR #1734, opened 2025-12-02
(Refactor) Extract file-specific information from book

piffio:
This PR implements the full refactoring suggested in #489 as a way to enable full multi-format support for all types of user interactions.

The PR is supposed to be 100% functionally compatible with the current codebase, i.e., it does not introduce any intended behavior change.

The only caveat is that it introduces the concept of "Primary File" which for now is simply the "first" file attached to a book. The idea is to follow up with subsequent (and smaller) PRs to implement a more flexible support for multiple formats per book, allowing the user to specify their preferred formats, etc.

Specifically, there are also a few corner cases I've noticed with the current implementation of "Additional Files" that can cause inconsistent states. I deliberately chose to not include the fix for those here to keep the PR "small" and avoid introducing unwanted regressions.

The aim is to make it easier to:
* Improve support for books with multiple file formats (PDF, EPUB, etc) for all interactions (Read, Download, etc)
* Support for merging/unmerging books
* Add support for additional file types
* Specify preferred formats at the user level

This is what the ER diagram for `book` and `book_file` would look like after the changes:

Screenshot 2025-12-02 at 23 00 51

And here is an example of the table content after data migration

The `book` table
Screenshot 2025-12-02 at 23 04 23

The `book_file` table (with `book_type` using the proper string value for enums)

Screenshot 2025-12-02 at 23 04 56

This PR introduces a lot of changes.

I was able to test most of them with book operations and OPDS, but unfortunately I'm unable to test the Kobo integration beyond what's covered by unit tests. 

@acx10 I do recommend giving it a spin, ideally migrating an existing library with a decent amount of books to ensure everything falls in the right place after the Flyway migration.

My suggestion would be to merge the PR to avoid further merge conflicts in the future and treat regressions as fixes that I'd be happy to work on one by one.

Given the size of the PR, though, I'd expect some additional scrutiny on the changes introduced :) 
@balazs-szucs your comments are also more than welcome.

## Detailed summary of changes

* [New data model for multi-format] Introduces `book_file` as the canonical place for file-specific fields (name/subpath, type, hashes, size, “isBook”/format flag), enabling multiple formats per book and separating “book metadata” from “file metadata”.
* [Flyway migration + data backfill] Adds/updates the migration that creates and backfills `book_file`, migrates data out of the legacy columns, and aligns enum storage (string values) with the new model.
[...]

PR #1879, opened 2025-12-14
feat(metadata) Save basic metadata to EPUB file - Bonus use Calibre Custom Columns for importing metadata.

### LINKED ISSUE #1878: [BUG] Importing CALIBRE processed (Polished) EPUB does not Extract ALL metadata into Booklore's Metadata fields.
# 🐛 Bug Report

## 📝 Description
Jumping from version v1.10.0 (I stayed there until OIDC was fixed) to v1.14.1. I can not get any of the metadata from Calibre processed EPUB's to populate Booklore's Metadata fields during "Bookdrop" import process (Same behavior on v1.10.1 but I Ignored it since I was NOT in the latest version).

The following metadata that exists in Calibre is not extracted to it's proper location on Booklore's side.

### Identifiers (_These are NOT extracted at all_):

- asin
- google
- goodreads
- hardcover
- comicvine

Calibre "**Tags**" are Mapped to Booklore's "**Genre**" section instead of Booklore's "**Tags**" section

### Working as expected:

- Title
- Language
- Authors
- Publisher
- Publish Date
- Series Name
- Series #
- ISBN <<-- Maps to ISBN-10 or ISBN-13 base on the count length of the **isbn** in Calibre

## 🔄 Steps to Reproduce

1. Process EPUB in Calibre and make sure ALL identifier Wanted, Tags, Series Name, Series Index is Set.
2. Polish Book just incase.
3. export book to Booklore's Bookdrop folder.
4. Import Book from Booklore's Bookdrop section in the UI.

**Result:** Results in Book imported with missing fields.

## ✅ Expected Behavior
All "CORE" fields that exist by default in Calibre to be mapped to Booklore's metadata fields

## 📸 Screenshots / Error Messages _(Optional)_

Image

## 💻 Environment

- **Booklore Version:** v1.14.1
- **OS:** UnRaid v7.2.2
- **Browser:** Brave 1.85.116
- **Installation:** UnRaid - Docker
- **Storage Type:** Synology NAS, SMB Share

## 💡 Possible Solution _(Optional)_
Update "EpubMetadataExtractor.java" to parse EPUB for Calibre "CORE" tags, and Allow the use of Calibre "Custom Columns" to map the rest of Booklore's metadata fields such as:

- Subtitle
- Genre
- Moods
- Series Total
- Pages
- Amazon Rating
- Amazon Review Count
- Goodreads Rating
- Goodreads Review Count
- Hardcover Rating
- Hardcover Review Count

You should end up importing the same book with all fields populated (Ebook in Image has all fields set in Calibre "CORE" fields, and Calibre Custom Columns)

### Custom Column lookup names:

- #subtitle
- #categories <--- This maps to "Genres" in the UI
- #moods
- #series_total
- #page_count
- #amazon_rating
- #amazon_review_count
- #goodreads_rating
- #goodreade_review_count
- #hardcover_rating
- #hardcover_review_count
- #isbn_10
- #isbn_13

Image

jadehawk:
# 🚀 Pull Request

## 📝 Description
Currently any EPUBS added to Booklore that were processed by Calibre will have the following Metadata  extracted and placed in the appropriate Booklore's Metadata Fields:

- Title
- Language
- Authors
- Publisher
- Publish Date
- Genres
- Series Name
- Series #
- ISBN
- Comment/Description

What fails to be imported from Calibre are the following:

### Identifiers:

- asin / amazon / mobi-asin
- goodreads
- google
- hardcover
- comicvine

Besides the basic metadata, everything else only lives in Booklore's Database, In the event your DB get corrupted or we migrate to another DB, or you simply want to start from scratch, none of the rich metadata will be preserved.

Fixes #1878

## 🛠️ Changes Implemented
With the implemented fix on this PR, we can now populate ALL metadata from Calibre (Polished) EPUBS and we also can save the extra (ALL) metadata we can see on Booklore's UI into the EPUB file, and Booklore will be able to extract the metadata.

The Following custom columns with their assigned "lookup" names" will be detected and placed in the proper Booklore's metadata fields:

<table>
  <tr>
    <th></th>
    <th>Calibre Custom Column (#lookupname)</th>
    <th>Booklore's Metadata Field</th>
  </tr>

  <tr>
    <td></td>
    <td>Subtitle (<strong>#subtitle</strong>)</td>
    <td>Subtitle</td>
  </tr>

  <tr>
    <td></td>
    <td>Moods (<strong>#moods</strong>)</td>
    <td>Moods</td>
  </tr>

  <tr>
    <td></td>
    <td>Tags (<strong>#extra_tags</strong>)</td>
    <td>Tags</td>
  </tr>

  <tr>
    <td></td>
    <td>Series Total (<strong>#series_total</strong>)</td>
    <td>Series Total</td>
  </tr>

  <tr>
    <td></td>
    <td>Pages (<strong>#pagecount</strong>)</td>
    <td>Pages</td>
  </tr>

  <tr>
    <td></td>
    <td>Amazon Rating (<strong>#amazon_rating</strong>)</td>
    <td>Amazon Rating</td>
  </tr>

  <tr>
    <td></td>
    <td>Amazon Reviews (<strong>#amazon_review_count</strong>)</td>
    <td>Amazon Reviews</td>
  </tr>

  <tr>
    <td></td>
    <td>Goodreads Rating (<strong>#goodreads_rating</strong>)</td>
    <td>Goodreads Rating</td>
  </tr>

  <tr>
    <td></td>
    <td>Goodreads Reviews (<strong>#goodreads_review_count</strong>)</td>
    <td>Goodreads Reviews</td>
  </tr>

  <tr>
    <td></td>
    <td>Hardcover Rating (<strong>#hardcover_rating</strong>)</td>
    <td>Hardcover Rating</td>
  </tr>

  <tr>
    <td></td>
    <td>Hardcover Reviews (<strong>#hardcover_review_count</strong>)</td>
    <td>Hardcover Reviews</td>
  </tr>
</table>

<br>

<table>
  <tr>
    <th></th>
    <th>Identifier</th>
    <th>In Calibre Identifier Section</th>
    <th>Booklore's Metadata Field</th>
  </tr>
  <tr>
    <td>1</td>
    <td>ISBN</td>
    <td>isbn:XXXXXXXX</td>
    <td>ISBN-10 or ISBN-13</td>
  </tr>
  <tr>
    <td></td>
    <td>ASIN</td>
    <td>asin:XXXXXXXX</td>
    <td>Amazon ASIN</td>
  </tr>
  <tr>
    <td></td>
    <td>AMAZON</td>
    <td>amazon:XXXXXXXX</td>
    <td>Amazon ASIN</td>
  </tr>
  <tr>
    <td></td>
    <td>MOBI-ASIN</td>
    <td>mobi-asin:XXXXXXXX</td>
    <td>Amazon ASIN</td>
  </tr>
  <tr>
    <td></td>
    <td>GOODREADS</td>
    <td>goodreads:XXXXXXXX</td>
    <td>Goodreads ID</td>
  </tr>
  <tr>
    <td></td>
    <td>GOOGLE</td>
    <td>google:XXXXXXXX</td>
    <td>Google Books ID</td>
  </tr>
  <tr>
    <td></td>
    <td>HARDCOVER</td>
    <td>hardcover:XXXXXXXX</td>
    <td>Hardcover ID</td>
  </tr>
  <tr>
    <td></td>
    <td>HARDCOVER_BOOK</td>
    <td>hardcover_book:XXXXXXXX</td>
    <td>Hardcover Book ID</td>
  </tr>
  <tr>
    <td></td>
    <td>COMICVINE</td>
    <td>comicvine:XXXXXXXX</td>
    <td>Comicvine ID</td>
  </tr>
  <tr>
    <td></td>
    <td>LUBIMYCZYTAC</td>
    <td>Lubimyczytac:XXXXXXXX</td>
    <td>LC ID</td>
  </tr>
  <tr>
    <td></td>
    <td>RANOBEDB</td>
    <td>ranobedb:XXXXXXXX</td>
    <td>Ranobedb ID</td>
  </tr>
</table>
<br>

**<sup>1</sup> Special Case:**
Calibre only accepts "1" isbn, Booklore will process this `isbn` and based on character count, will place it in ISBN-10 or ISBN-13.

## Enhancements

While working on this PR, I realized that most of Booklore's Rich Metadata (Fetched from metadata providers) lives only on the database, in the event user wants to start from scratch or database gets corrupted, user will need to fetch all the metadata again.
To fix this, we will now embed all Booklore's metadata inside the EPUB, and next time you IMPORT the EPUB, Booklore will 
recognize the embedded metadata and populate the proper fields.

The following additional metadata will be saved to the EPUB file together with the standard EPUB 3 fields, and Identifiers:

<table>
  <tr>
    <th></th>
    <th>Booklore Metadata</th>
    <th>Tag</th>
  </tr>
  <tr>
    <td></td>
    <td>Moods</td>
    <td>booklore:moods</td>
  </tr>
  <tr>
    <td></td>
    <td>Tags</td>
    <td>booklore:tags</td>
  </tr>
  <tr>
    <td></td>
    <td>Series Total</td>
    <td>booklore:series_total</td>
  </tr>
  <tr>
    <td></td>
    <td>Pages</td>
    <td>booklore:page_count</td>
  </tr>
  <tr>
    <td></td>
    <td>Amazon Rating</td>
    <td>booklore:amazon_rating</td>
  </tr>
  <tr>
    <td></td>
    <td>Amazon Reviews</td>
    <td>booklore:amazon_review_count</td>
  </tr>
  <tr>
    <td></td>
    <td>Goodreads Rating</td>
    <td>booklore:goodreads_rating</td>
  </tr>
  <tr>
    <td></td>
    <td>Goodreads Reviews</td>
    <td>booklore:goodreads_review_count</td>
  </tr>
  <tr>
    <td></td>
    <td>Hardcover Rating</td>
    <td>booklore:hardcover_rating</td>
  </tr>
  <tr>
    <td></td>
    <td>Hardcover Reviews</td>
    <td>booklore:hardcover_review_count</td>
  </tr>
  <tr>
    <td></td>
    <td>Lubimyczytac Rating</td>
    <td>booklore:lubimyczytac_rating</td>
  </tr>
  <tr>
    <td></td>
    <td>Ranobedb Rating</td>
    <td>booklore:ranobedb_rating</td>
  </tr>
</table>
<br>

## 🧪 Testing Strategy
Docker image for testing with Fix applied can be downloaded here:
``` code
docker pull jadehawk/jh-booklore:1.18.5-metadata-saver
```
Also added extra JUNIT test so we can test extracting all the custom columns using **"gradlew Test"**, all current test pass.

- [ ] Flyway migration versioning follows correct sequence _(if database schema modified)_

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/config/security/filter/AudiobookStreamingJwtFilter.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/config/security/filter/AudiobookStreamingJwtFilter.java
@@ -24,11 +24,6 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.regex.Pattern;
 
-/**
- * JWT filter for audiobook streaming endpoints that supports both:
- * 1. Authorization header (Bearer token) - for fetch() requests
- * 2. Query parameter (token) - for HTML5 audio element compatibility
- */
 @Component
 @AllArgsConstructor
 public class AudiobookStreamingJwtFilter extends OncePerRequestFilter {
@@ -52,7 +47,6 @@ public class AudiobookStreamingJwtFilter extends OncePerRequestFilter {
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
             throws ServletException, IOException {
 
-        // Try Authorization header first, then fall back to query parameter
         String token = extractTokenFromHeader(request);
         if (token == null) {
             token = request.getParameter("token");
--- a/booklore-api/src/main/java/com/adityachandel/booklore/config/security/filter/EpubStreamingJwtFilter.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/config/security/filter/EpubStreamingJwtFilter.java
@@ -24,11 +24,6 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.regex.Pattern;
 
-/**
- * JWT filter for EPUB streaming endpoints that supports both:
- * 1. Authorization header (Bearer token) - for fetch() requests
- * 2. Query parameter (token) - for browser-initiated requests (fonts, images in CSS)
- */
 @Component
 @AllArgsConstructor
 public class EpubStreamingJwtFilter extends OncePerRequestFilter {
@@ -43,7 +38,6 @@ public class EpubStreamingJwtFilter extends OncePerRequestFilter {
     @Override
     protected boolean shouldNotFilter(HttpServletRequest request) {
         String path = request.getRequestURI();
-        // Only filter requests to EPUB file streaming endpoint
         return !EPUB_STREAMING_ENDPOINT_PATTERN.matcher(path).matches();
     }
 
@@ -51,7 +45,6 @@ public class EpubStreamingJwtFilter extends OncePerRequestFilter {
     protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
             throws ServletException, IOException {
 
-        // Try Authorization header first, then fall back to query parameter
         String token = extractTokenFromHeader(request);
         if (token == null) {
             token = request.getParameter("token");
--- a/booklore-api/src/main/java/com/adityachandel/booklore/controller/UserStatsController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/controller/UserStatsController.java
@@ -31,6 +31,20 @@ public class UserStatsController {
         return ResponseEntity.ok(heatmapData);
     }
 
+    @Operation(summary = "Get reading session heatmap for a month", description = "Returns daily reading session counts for the authenticated user for a specific year and month")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Heatmap data retrieved successfully"),
+            @ApiResponse(responseCode = "401", description = "Unauthorized")
+    })
+    @GetMapping("/heatmap/monthly")
+    @PreAuthorize("@securityUtil.canAccessUserStats() or @securityUtil.isAdmin()")
+    public ResponseEntity<List<ReadingSessionHeatmapResponse>> getHeatmapForMonth(
+            @RequestParam int year,
+            @RequestParam int month) {
+        List<ReadingSessionHeatmapResponse> heatmapData = readingSessionService.getSessionHeatmapForMonth(year, month);
+        return ResponseEntity.ok(heatmapData);
+    }
+
     @Operation(summary = "Get reading session timeline for a week", description = "Returns reading sessions grouped by book for calendar timeline view")
     @ApiResponses({
             @ApiResponse(responseCode = "200", description = "Timeline data retrieved successfully"),
@@ -108,5 +122,17 @@ public class UserStatsController {
     public ResponseEntity<List<CompletionTimelineResponse>> getCompletionTimeline(@RequestParam int year) {
         List<CompletionTimelineResponse> timeline = readingSessionService.getCompletionTimeline(year);
         return ResponseEntity.ok(timeline);
+    }
+
+    @Operation(summary = "Get book completion heatmap", description = "Returns monthly book completion counts for the last 10 years for the authenticated user")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Book completion heatmap data retrieved successfully"),
+            @ApiResponse(responseCode = "401", description = "Unauthorized")
+    })
+    @GetMapping("/book-completion-heatmap")
+    @PreAuthorize("@securityUtil.canAccessUserStats() or @securityUtil.isAdmin()")
+    public ResponseEntity<List<BookCompletionHeatmapResponse>> getBookCompletionHeatmap() {
+        List<BookCompletionHeatmapResponse> heatmapData = readingSessionService.getBookCompletionHeatmap();
+        return ResponseEntity.ok(heatmapData);
     }
 }
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/controller/MobileBookController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/controller/MobileBookController.java
@@ -1,0 +1,143 @@
+package com.adityachandel.booklore.mobile.controller;
+
+import com.adityachandel.booklore.mobile.dto.*;
+import com.adityachandel.booklore.mobile.service.MobileBookService;
+import com.adityachandel.booklore.model.enums.ReadStatus;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import jakarta.validation.Valid;
+import lombok.AllArgsConstructor;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+
+@AllArgsConstructor
+@RestController
+@RequestMapping("/api/mobile/v1/books")
+@Tag(name = "Mobile Books", description = "Mobile-optimized endpoints for book operations")
+public class MobileBookController {
+
+    private final MobileBookService mobileBookService;
+
+    @Operation(summary = "Get paginated book list",
+            description = "Retrieve a paginated list of books with optional filtering by library, shelf, status, and search text.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Books retrieved successfully"),
+            @ApiResponse(responseCode = "403", description = "Access denied")
+    })
+    @GetMapping
+    public ResponseEntity<MobilePageResponse<MobileBookSummary>> getBooks(
+            @Parameter(description = "Page number (0-indexed)") @RequestParam(required = false, defaultValue = "0") Integer page,
+            @Parameter(description = "Page size (max 50)") @RequestParam(required = false, defaultValue = "20") Integer size,
+            @Parameter(description = "Sort field (title, addedOn, lastReadTime, seriesName)") @RequestParam(required = false, defaultValue = "addedOn") String sort,
+            @Parameter(description = "Sort direction (asc, desc)") @RequestParam(required = false, defaultValue = "desc") String dir,
+            @Parameter(description = "Filter by library ID") @RequestParam(required = false) Long libraryId,
+            @Parameter(description = "Filter by shelf ID") @RequestParam(required = false) Long shelfId,
+            @Parameter(description = "Filter by read status") @RequestParam(required = false) ReadStatus status,
+            @Parameter(description = "Search in title, author, series") @RequestParam(required = false) String search) {
+
+        return ResponseEntity.ok(mobileBookService.getBooks(
+                page, size, sort, dir, libraryId, shelfId, status, search));
+    }
+
+    @Operation(summary = "Get book details",
+            description = "Retrieve full details for a specific book.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Book details retrieved successfully"),
+            @ApiResponse(responseCode = "403", description = "Access denied"),
+            @ApiResponse(responseCode = "404", description = "Book not found")
+    })
+    @GetMapping("/{bookId}")
+    public ResponseEntity<MobileBookDetail> getBookDetail(
+            @Parameter(description = "Book ID") @PathVariable Long bookId) {
+
+        return ResponseEntity.ok(mobileBookService.getBookDetail(bookId));
+    }
+
+    @Operation(summary = "Search books",
+            description = "Search books by query text in title, author, and series name.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Search results retrieved successfully"),
+            @ApiResponse(responseCode = "400", description = "Query parameter required")
+    })
+    @GetMapping("/search")
+    public ResponseEntity<MobilePageResponse<MobileBookSummary>> searchBooks(
+            @Parameter(description = "Search query", required = true) @RequestParam String q,
+            @Parameter(description = "Page number (0-indexed)") @RequestParam(required = false, defaultValue = "0") Integer page,
+            @Parameter(description = "Page size (max 50)") @RequestParam(required = false, defaultValue = "20") Integer size) {
+
+        return ResponseEntity.ok(mobileBookService.searchBooks(q, page, size));
+    }
+
+    @Operation(summary = "Get continue reading list",
+            description = "Get books currently in progress, sorted by last read time (most recent first).")
+    @ApiResponse(responseCode = "200", description = "Continue reading list retrieved successfully")
+    @GetMapping("/continue-reading")
+    public ResponseEntity<List<MobileBookSummary>> getContinueReading(
+            @Parameter(description = "Maximum number of books to return") @RequestParam(required = false, defaultValue = "10") Integer limit) {
+
+        return ResponseEntity.ok(mobileBookService.getContinueReading(limit));
+    }
+
+    @Operation(summary = "Get recently added books",
+            description = "Get books added in the last 30 days, sorted by added date (most recent first).")
+    @ApiResponse(responseCode = "200", description = "Recently added books retrieved successfully")
+    @GetMapping("/recently-added")
+    public ResponseEntity<List<MobileBookSummary>> getRecentlyAdded(
+            @Parameter(description = "Maximum number of books to return") @RequestParam(required = false, defaultValue = "10") Integer limit) {
+
+        return ResponseEntity.ok(mobileBookService.getRecentlyAdded(limit));
+    }
+
+    @Operation(summary = "Update book read status",
+            description = "Update the read status for a book.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Status updated successfully"),
+            @ApiResponse(responseCode = "403", description = "Access denied"),
+            @ApiResponse(responseCode = "404", description = "Book not found")
+    })
+    @PutMapping("/{bookId}/status")
+    public ResponseEntity<Void> updateStatus(
+            @Parameter(description = "Book ID") @PathVariable Long bookId,
+            @Valid @RequestBody UpdateStatusRequest request) {
+
+        mobileBookService.updateReadStatus(bookId, request.getStatus());
+        return ResponseEntity.ok().build();
+    }
+
+    @Operation(summary = "Update book personal rating",
+            description = "Update the personal rating for a book (1-5).")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Rating updated successfully"),
+            @ApiResponse(responseCode = "403", description = "Access denied"),
+            @ApiResponse(responseCode = "404", description = "Book not found")
+    })
+    @PutMapping("/{bookId}/rating")
+    public ResponseEntity<Void> updateRating(
+            @Parameter(description = "Book ID") @PathVariable Long bookId,
+            @Valid @RequestBody UpdateRatingRequest request) {
+
+        mobileBookService.updatePersonalRating(bookId, request.getRating());
+        return ResponseEntity.ok().build();
+    }
+
+    @Operation(summary = "Get books by magic shelf",
+            description = "Retrieve a paginated list of books matching a magic shelf's rules.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "Books retrieved successfully"),
+            @ApiResponse(responseCode = "403", description = "Access denied"),
+            @ApiResponse(responseCode = "404", description = "Magic shelf not found")
+    })
+    @GetMapping("/magic-shelf/{magicShelfId}")
+    public ResponseEntity<MobilePageResponse<MobileBookSummary>> getBooksByMagicShelf(
+            @Parameter(description = "Magic shelf ID") @PathVariable Long magicShelfId,
+            @Parameter(description = "Page number (0-indexed)") @RequestParam(required = false, defaultValue = "0") Integer page,
+            @Parameter(description = "Page size (max 50)") @RequestParam(required = false, defaultValue = "20") Integer size) {
+
+        return ResponseEntity.ok(mobileBookService.getBooksByMagicShelf(magicShelfId, page, size));
+    }
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/controller/MobileLibraryController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/controller/MobileLibraryController.java
@@ -1,0 +1,60 @@
+package com.adityachandel.booklore.mobile.controller;
+
+import com.adityachandel.booklore.config.security.service.AuthenticationService;
+import com.adityachandel.booklore.mobile.dto.MobileLibrarySummary;
+import com.adityachandel.booklore.mobile.mapper.MobileBookMapper;
+import com.adityachandel.booklore.model.dto.BookLoreUser;
+import com.adityachandel.booklore.model.dto.Library;
+import com.adityachandel.booklore.model.entity.LibraryEntity;
+import com.adityachandel.booklore.repository.BookRepository;
+import com.adityachandel.booklore.repository.LibraryRepository;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import lombok.AllArgsConstructor;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@AllArgsConstructor
+@RestController
+@RequestMapping("/api/mobile/v1/libraries")
+@Tag(name = "Mobile Libraries", description = "Mobile-optimized endpoints for library operations")
+public class MobileLibraryController {
+
+    private final AuthenticationService authenticationService;
+    private final LibraryRepository libraryRepository;
+    private final BookRepository bookRepository;
+    private final MobileBookMapper mobileBookMapper;
+
+    @Operation(summary = "Get user's accessible libraries",
+            description = "Retrieve a list of libraries the current user has access to.")
+    @ApiResponse(responseCode = "200", description = "Libraries retrieved successfully")
+    @GetMapping
+    public ResponseEntity<List<MobileLibrarySummary>> getLibraries() {
+        BookLoreUser user = authenticationService.getAuthenticatedUser();
+
+        List<LibraryEntity> libraries;
+        if (user.getPermissions().isAdmin()) {
+            libraries = libraryRepository.findAll();
+        } else {
+            List<Long> libraryIds = user.getAssignedLibraries() != null
+                    ? user.getAssignedLibraries().stream().map(Library::getId).collect(Collectors.toList())
+                    : List.of();
+            libraries = libraryRepository.findByIdIn(libraryIds);
+        }
+
+        List<MobileLibrarySummary> summaries = libraries.stream()
+                .map(library -> {
+                    long bookCount = bookRepository.countByLibraryId(library.getId());
+                    return mobileBookMapper.toLibrarySummary(library, bookCount);
+                })
+                .collect(Collectors.toList());
+
+        return ResponseEntity.ok(summaries);
+    }
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/controller/MobileShelfController.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/controller/MobileShelfController.java
@@ -1,0 +1,90 @@
+package com.adityachandel.booklore.mobile.controller;
+
+import com.adityachandel.booklore.config.security.service.AuthenticationService;
+import com.adityachandel.booklore.mobile.dto.MobileMagicShelfSummary;
+import com.adityachandel.booklore.mobile.dto.MobileShelfSummary;
+import com.adityachandel.booklore.mobile.mapper.MobileBookMapper;
+import com.adityachandel.booklore.model.dto.BookLoreUser;
+import com.adityachandel.booklore.model.entity.MagicShelfEntity;
+import com.adityachandel.booklore.model.entity.ShelfEntity;
+import com.adityachandel.booklore.repository.MagicShelfRepository;
+import com.adityachandel.booklore.repository.ShelfRepository;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import lombok.AllArgsConstructor;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@AllArgsConstructor
+@RestController
+@RequestMapping("/api/mobile/v1/shelves")
+@Tag(name = "Mobile Shelves", description = "Mobile-optimized endpoints for shelf operations")
+public class MobileShelfController {
+
+    private final AuthenticationService authenticationService;
+    private final ShelfRepository shelfRepository;
+    private final MagicShelfRepository magicShelfRepository;
+    private final MobileBookMapper mobileBookMapper;
+
+    @Operation(summary = "Get user's shelves",
+            description = "Retrieve a list of shelves the current user owns or are public.")
+    @ApiResponse(responseCode = "200", description = "Shelves retrieved successfully")
+    @GetMapping
+    public ResponseEntity<List<MobileShelfSummary>> getShelves() {
+        BookLoreUser user = authenticationService.getAuthenticatedUser();
+        Long userId = user.getId();
+
+        List<ShelfEntity> shelves = shelfRepository.findByUserIdOrPublicShelfTrue(userId);
+
+        List<MobileShelfSummary> summaries = shelves.stream()
+                .map(mobileBookMapper::toShelfSummaryFromEntity)
+                .collect(Collectors.toList());
+
+        return ResponseEntity.ok(summaries);
+    }
+
+    @Operation(summary = "Get user's magic shelves",
+            description = "Retrieve a list of magic shelves the current user owns or are public.")
+    @ApiResponse(responseCode = "200", description = "Magic shelves retrieved successfully")
+    @GetMapping("/magic")
+    public ResponseEntity<List<MobileMagicShelfSummary>> getMagicShelves() {
+        BookLoreUser user = authenticationService.getAuthenticatedUser();
+        Long userId = user.getId();
+
+        // Get user's own magic shelves
+        List<MagicShelfEntity> userShelves = magicShelfRepository.findAllByUserId(userId);
+
+        // Get public magic shelves
+        List<MagicShelfEntity> publicShelves = magicShelfRepository.findAllByIsPublicIsTrue();
+
+        // Combine and deduplicate (user's shelves that are also public shouldn't appear twice)
+        Set<Long> seenIds = new HashSet<>();
+        List<MagicShelfEntity> allShelves = new ArrayList<>();
+
+        for (MagicShelfEntity shelf : userShelves) {
+            if (seenIds.add(shelf.getId())) {
+                allShelves.add(shelf);
+            }
+        }
+        for (MagicShelfEntity shelf : publicShelves) {
+            if (seenIds.add(shelf.getId())) {
+                allShelves.add(shelf);
+            }
+        }
+
+        List<MobileMagicShelfSummary> summaries = allShelves.stream()
+                .map(mobileBookMapper::toMagicShelfSummary)
+                .collect(Collectors.toList());
+
+        return ResponseEntity.ok(summaries);
+    }
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileBookDetail.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileBookDetail.java
@@ -1,0 +1,83 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.util.List;
+import java.util.Set;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobileBookDetail {
+    private Long id;
+    private String title;
+    private List<String> authors;
+    private String thumbnailUrl;
+    private String readStatus;
+    private Integer personalRating;
+    private String seriesName;
+    private Float seriesNumber;
+    private Long libraryId;
+    private Instant addedOn;
+    private Instant lastReadTime;
+
+    private String subtitle;
+    private String description;
+    private Set<String> categories;
+    private String publisher;
+    private LocalDate publishedDate;
+    private Integer pageCount;
+    private String isbn13;
+    private String language;
+    private Double goodreadsRating;
+    private Integer goodreadsReviewCount;
+    private String libraryName;
+    private List<MobileShelfSummary> shelves;
+    private Float readProgress;
+    private String primaryFileType;
+    private List<String> fileTypes;
+    private List<MobileBookFile> files;
+
+    private EpubProgress epubProgress;
+    private PdfProgress pdfProgress;
+    private CbxProgress cbxProgress;
+    private AudiobookProgress audiobookProgress;
+
+    @Data
+    @Builder
+    public static class EpubProgress {
+        private String cfi;
+        private String href;
+        private Float percentage;
+        private Instant updatedAt;
+    }
+
+    @Data
+    @Builder
+    public static class PdfProgress {
+        private Integer page;
+        private Float percentage;
+        private Instant updatedAt;
+    }
+
+    @Data
+    @Builder
+    public static class CbxProgress {
+        private Integer page;
+        private Float percentage;
+        private Instant updatedAt;
+    }
+
+    @Data
+    @Builder
+    public static class AudiobookProgress {
+        private Long positionMs;
+        private Integer trackIndex;
+        private Float percentage;
+        private Instant updatedAt;
+    }
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileBookFile.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileBookFile.java
@@ -1,0 +1,16 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobileBookFile {
+    private Long id;
+    private String bookType;
+    private Long fileSizeKb;
+    private String fileName;
+    private boolean isPrimary;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileBookSummary.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileBookSummary.java
@@ -1,0 +1,25 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+import java.time.Instant;
+import java.util.List;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobileBookSummary {
+    private Long id;
+    private String title;
+    private List<String> authors;
+    private String thumbnailUrl;
+    private String readStatus;
+    private Integer personalRating;
+    private String seriesName;
+    private Float seriesNumber;
+    private Long libraryId;
+    private Instant addedOn;
+    private Instant lastReadTime;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileLibrarySummary.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileLibrarySummary.java
@@ -1,0 +1,15 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobileLibrarySummary {
+    private Long id;
+    private String name;
+    private String icon;
+    private long bookCount;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileMagicShelfSummary.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileMagicShelfSummary.java
@@ -1,0 +1,16 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobileMagicShelfSummary {
+    private Long id;
+    private String name;
+    private String icon;
+    private String iconType;
+    private boolean publicShelf;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobilePageResponse.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobilePageResponse.java
@@ -1,0 +1,33 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobilePageResponse<T> {
+    private List<T> content;
+    private int page;
+    private int size;
+    private long totalElements;
+    private int totalPages;
+    private boolean hasNext;
+    private boolean hasPrevious;
+
+    public static <T> MobilePageResponse<T> of(List<T> content, int page, int size, long totalElements) {
+        int totalPages = size > 0 ? (int) Math.ceil((double) totalElements / size) : 0;
+        return MobilePageResponse.<T>builder()
+                .content(content)
+                .page(page)
+                .size(size)
+                .totalElements(totalElements)
+                .totalPages(totalPages)
+                .hasNext(page < totalPages - 1)
+                .hasPrevious(page > 0)
+                .build();
+    }
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileShelfSummary.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/MobileShelfSummary.java
@@ -1,0 +1,16 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class MobileShelfSummary {
+    private Long id;
+    private String name;
+    private String icon;
+    private int bookCount;
+    private boolean publicShelf;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/UpdateRatingRequest.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/UpdateRatingRequest.java
@@ -1,0 +1,12 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import jakarta.validation.constraints.Max;
+import jakarta.validation.constraints.Min;
+import lombok.Data;
+
+@Data
+public class UpdateRatingRequest {
+    @Min(value = 1, message = "Rating must be at least 1")
+    @Max(value = 5, message = "Rating must be at most 5")
+    private Integer rating;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/UpdateStatusRequest.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/dto/UpdateStatusRequest.java
@@ -1,0 +1,11 @@
+package com.adityachandel.booklore.mobile.dto;
+
+import com.adityachandel.booklore.model.enums.ReadStatus;
+import jakarta.validation.constraints.NotNull;
+import lombok.Data;
+
+@Data
+public class UpdateStatusRequest {
+    @NotNull(message = "Status is required")
+    private ReadStatus status;
+}
--- a/booklore-api/src/main/java/com/adityachandel/booklore/mobile/mapper/MobileBookMapper.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/mobile/mapper/MobileBookMapper.java
@@ -1,0 +1,284 @@
+package com.adityachandel.booklore.mobile.mapper;
+
+import com.adityachandel.booklore.mobile.dto.MobileBookDetail;
+import com.adityachandel.booklore.mobile.dto.MobileBookFile;
+import com.adityachandel.booklore.mobile.dto.MobileBookSummary;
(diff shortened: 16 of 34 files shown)


.dockerignore
```diff
@@ -0,0 +1,34 @@
...
+out/
+
+# Testing and discovery scripts
+check_calibre_fields.py
+example.epub
+
+# Dev config
+booklore-api/src/main/resources/application-local.yaml
+local/
+shared/
```
balazs-szucs: Unrelated change

.gitignore
```diff
@@ -43,4 +43,10 @@ local/
...
 booklore-api/src/main/resources/application-local.yaml
-/shared/
\ No newline at end of file
+/shared/
+
+### Testing and discovery scripts ###
+check_calibre_fields.py
+
+### Reference Documentation ###
+CALIBRE_FIELD_MAPPING_REFERENCE.txt
```
balazs-szucs: Seems also unrelated?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -35,6 +54,103 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
...
+        // ========================================================================
+        // NOTE: Set<String> fields (categories, tags, moods, authors) are handled
+        // separately in the extractCalibreUserMetadata() method as they require
+        // special processing for JSON arrays and comma-separated values
+        // ========================================================================
+    );
+    
+    // ============================================================================
+    // END OF CALIBRE USER METADATA FIELD MAPPING SYSTEM
+    // ============================================================================
```
balazs-szucs: These comments unnecessary, decrease readability.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -35,6 +54,103 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
...
+     */
+    /**
+     * Pre-built mapping for ALL Known (As of 12/14/2025) BookMetadata fields from CALIBRE_FIELD_MAPPING_REFERENCE.txt.
+     * Users can create Calibre custom columns with these exact names and they will be automatically extracted.
+     * No code changes needed - just create the custom column in Calibre with the correct name!
+     * 
+     * IMPORTANT: Calibre requires custom column lookup names to be LOWERCASE.
+     * This mapping includes both camelCase and lowercase versions for compatibility.
+     */
+    private static final Map<String, BiConsumer<BookMetadata.BookMetadataBuilder, String>> CALIBRE_FIELD_MAPPINGS = Map.ofEntries(
```
balazs-szucs, line 84: JUnit tests for this, perhaps? And other methods also obviously.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -173,32 +312,157 @@ public BookMetadata extractMetadata(File epubFile) {
...
+                                        } 
+                                        // Check for ISBN-10 (10 digits or 9 digits + X)
+                                        else if (cleanValue.length() == 10 && cleanValue.matches("\\d{10}|[0-9]{9}[xX]") && builderMeta.build().getIsbn10() == null) {
+                                            builderMeta.isbn10(cleanValue);
+                                            log.debug("IDENTIFIER: Set ISBN-10 from length detection (10 chars): '{}'", cleanValue);
+                                        } 
+                                        // If no match, log for debugging
+                                        else {
+                                            log.trace("IDENTIFIER: No standard format matched - cleaned value: '{}', length: {}, is digits: {}", 
+                                                cleanValue, cleanValue.length(), cleanValue.matches("\\d+"));
```
balazs-szucs: This block needs significant refactoring:
- DRY Violation, the identifier extraction logic is repeated multiple times. We should extract this into a helper method (e.g., processIdentifier(prefix, setter)) to handle the null checks and logging centrally.
- We are parsing new JSONObject(content) multiple times for the same property. Please parse this once at the top of the block and reuse the object.
- The pattern builderMeta.build().getField() inside the loop is the "Build-to-Check" anti-pattern. It creates a new immutable object on every iteration just to check a null state. Please introduce a Set<String> processedFields or add getters to the Builder itself to avoid this allocation.
- Using String.matches() inside a loop recompiles the Regex pattern every time. Please move these to static final Pattern constants.
- Brittle Code: There are several "magic numbers" (e.g., substring(5), substring(10)). These should be derived from the prefix length (prefix.length()) to prevent off-by-one errors if prefixes change.

jadehawk:
@balazs-szucs I've addressed all the review feedback:

- Eliminated DRY violations with helper methods for identifier extraction
- Fixed Build-to-Check anti-pattern using processedFields Set tracking
- Moved regex patterns to static constants to avoid recompilation
- Replaced magic numbers with calculated prefix lengths
- Consolidated repetitive Set field and pagecount parsing patterns
- Removed excessive documentation and comments
- Removed unrelated file changes (.dockerignore and .gitignore entries)

The refactoring reduces the code by 207 lines (129 insertions, 336 deletions) while maintaining all functionality. Code has been tested and works without issues. Ready for re-review.

balazs-szucs:
@jadehawk Thanks! Sorry re-reading my review it is a bit harsh, apologies for that. Many thanks for your work and the PR!

jadehawk:
@balazs-szucs 

I got thick skin, so no worries on how the tone of the comment was, did not bother me in any way.

I've added plenty of JUnit tests for the EpubMetadataExtractor class. All tests are passing successfully (verified in build/reports/tests) results generated on that folder. The tests cover the metadata extraction scenarios and including identifier handling, title extraction, author parsing, and hope I cover every angle there. Ready for your review!

Thanx again for this excellent piece of software.

P.S. TEST Docker Image was rebuild and reuploaded.

jadehawk:
@balazs-szucs 
Rebased on latest dev - all JUnit tests passing. Ready for another tough review.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -173,33 +269,36 @@ public BookMetadata extractMetadata(File epubFile) {
...
+                                    try {
+                                        JSONObject jsonroot = new JSONObject(content);
+                                        extractCalibreUserMetadata(jsonroot, builderMeta, categories);
+                                    } catch (JSONException ignored) {
+                                    }
+                                }
                             }
                             case "creator" -> authors.add(text);
-                            case "subject" -> categories.add(text);
+                            case "subject" -> tags.add(text);
```
MightyHalfGinger: This will cause the epub writer to be a bit confused since it saves the categories list to the subject tag: https://github.com/booklore-app/booklore/blob/36d5f9e6e059b5fd864642354fba5b21cfff1297/booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriter.java#L100-L108

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -379,4 +596,60 @@ private byte[] extractFileFromZip(File epubFile, String path) {
...
+                        continue;
+                    }
+                    
+                    String value = String.valueOf(rawValue).trim();
+                    if (value.isEmpty() || "null".equals(value)) {
+                        continue;
+                    }
+                    
+                    if ("#categories".equals(fieldName)) {
+                        extractSetField(value, categories);
```
MightyHalfGinger: Could we use `extractAndSetUserMetadataSet(value, builder::categories);` here rather than needing to pass in the categories set from the caller?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
jadehawk: This was a section that cause confusion for me. In Calibre they are called "Tags" and when you want to organize by "Genres" a custom column is created; In booklore we have a placeholder for "Tags" and "Genres". So I opted to keep Tags from Calibre to "Tags" on Booklore. What I forgot to do was map the #genres custom column from Calibre
to Booklore "Genres" placeholder. What I can do is Populate "Genres" with the custom column value if it exists, if it does not populate "Genres" with the values from "Tags", another question will be what to do with the "Tags" placeholder in Booklore, If the "Tags" are used for "Genres", do we use them again for "Tags" on Booklore?

"Tags" can be anything really ("Norse God's", "Political", "Lost Dogs"...), you can go wild on them, "Genres" on the other hand IMHO are stricter ("Fiction", "Sci-fi", "History", "Religion"....) kind of how a library would organize the books sections.

jadehawk:
@MightyHalfGinger 

Ok, made the following changes.
If user has created a custom column called "genres" or "categories" and has a VALUE in that column then we map that value to the "Genres" in booklore.

If user has created a custom column called "genres" or "categories" and has a NO VALUE in that column then we map the value of "tags" to the "Genres".

If user does not have any #genres or #categories custom column, we map "tags" to "Genres" in booklore which is current booklore behavior.

Essentially, leaving the current behavior of Booklore mapping tags to genres, unless the user embeds the custom column to their EPUBs in Calibre. What do you think of this approach?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
MightyHalfGinger: I reopened calibre for the first time in a while and now I understand your confusion/concern since it doesn't have a genre section at all and the tags are used in place of that. I'm not sure what is "expected" for the genre vs tag fields in booklore so the maintainers will have to figure out how best to deal with this... I just was bringing up that if you saved some genres to the epub then re-imported the same book it would move them to the tags section with the old logic. My personal take would be to always lock the subject tag to genre as it is in the writer/previously, but so long as its consistent between the two it doesn't matter much to me.

Looking at the other extractors the tags field in booklore doesn't seem to ever be read in/populated from files; in the CBX file type we actually [combine both "Tags" and "Genre" data into just the `categories` variable](https://github.com/booklore-app/booklore/blob/97cb7812291a5d2d6f5e99b515a525f0925e7d5d/booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/CbxMetadataExtractor.java#L195-L196), which then maps to the genre field in the booklore UI IIRC.

I wonder if subjects should just be always mapped to genres always so that the read/write pair are consistent, and then let calibre's tags from the user_metadata go into the tags data on their own?

jadehawk:
> Could we use `extractAndSetUserMetadataSet(value, builder::categories);` here rather than needing to pass in the categories set from the caller?

@MightyHalfGinger

You raised a great question here. I did not use that function because of the way I was generating the "categories" value, and when using the builder, it puts the first value found when called and ignores the rest.

But this made me think of a user case where both __#genres__ AND #categories custom column exists and have values inside the EPUB metadata. To keep with how the rest of the metadata fields are extracted and set, I chose to prioritize __"#genres"__, this essentially allows me to use `extractAndSetUserMetadataSet(value, builder::categories);` to set the categories.

I refactored the code again to work with your new addition to the authors role extraction (PR #1987).

I noticed that most of the rich metadata collected in Booklore lives mostly in the DB. If this DB gets corrupted or if a user simply wants to start from scratch, all the metadata will need to be re-added, since it is not preserved in the EPUB.

I added provision to save metadata into the EPUB leveraging Calibre's custom columns. This way if a user re-adds an EPUB updated in Booklore to another Booklore instance, the metadata will be detected and extracted. And if they want to use Calibre, as long as the proper custom columns are in place, the metadata will be visible in Calibre as well, making it a more robust Calibre Integration.

Ready for review.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -184,9 +287,17 @@ public BookMetadata extractMetadata(File epubFile) {
...
+                                    try {
+                                        JSONObject jsonroot = new JSONObject(content);
+                                        extractCalibreUserMetadata(jsonroot, builderMeta, categories);
+                                    } catch (JSONException ignored) {
+                                    }
+                                }
                             }
                             case "creator" -> {
-                                String role = el.getAttributeNS(OPF_NS, "role");
+                                String role = el.getAttributeNS("http://www.idpf.org/2007/opf", "role");
```
MightyHalfGinger: Lets keep the constant for the namespace

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
 import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.regex.Pattern;
 
 @Slf4j
 @Component
 public class EpubMetadataExtractor implements FileMetadataExtractor {
 
     private static final Pattern YEAR_ONLY_PATTERN = Pattern.compile("^\\d{4}$");
-    private static final String OPF_NS = "http://www.idpf.org/2007/opf";
```
MightyHalfGinger, line 39: Please keep this constant

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -198,31 +309,32 @@ public BookMetadata extractMetadata(File epubFile) {
...
+                                    categories.add(text);
+                                }
+                            }
                             case "description" -> builderMeta.description(text);
                             case "publisher" -> builderMeta.publisher(text);
                             case "language" -> builderMeta.language(text);
                             case "identifier" -> {
-                                String scheme = el.getAttributeNS(OPF_NS, "scheme").toUpperCase();
-                                String value = text.toLowerCase().startsWith("isbn:") ? text.substring(5) : text;
+                                String scheme = el.getAttributeNS("http://www.idpf.org/2007/opf", "scheme").toUpperCase();
```
MightyHalfGinger: Lets keep the constant

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -362,8 +612,7 @@ private String findCoverImageHrefInOpf(File epubFile) {
                         String properties = item.getAttribute("properties");
                         if (properties != null && properties.contains("cover-image")) {
                             String href = item.getAttribute("href");
-                            String decodedHref = URLDecoder.decode(href, StandardCharsets.UTF_8);
-                            return resolvePath(opfPath, decodedHref);
+                            return resolvePath(opfPath, href);
```
MightyHalfGinger: Why did this get changed?

booklore-api/src/test/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriterTest.java
```diff
@@ -55,36 +52,53 @@ void setUp() {
...
-            existingMetadata.append("<dc:creator id=\"creator02\">Alice</dc:creator>");
-            existingMetadata.append("<meta property=\"role\" refines=\"#creator02\">ill</meta>");
-            existingMetadata.append("</metadata>");
-            String opfContent = String.format("""
-                <?xml version="1.0" encoding="UTF-8"?>
-                <package xmlns="http://www.idpf.org/2007/opf" version="3.0">
-                    %s
-                </package>
-                """, existingMetadata);
-            File epubFile = createEpubWithOpf(opfContent, "test-metadata-" + System.nanoTime() + ".epub");
```
MightyHalfGinger, line 96: This seems like a bad merge resolution since it removed the testcase

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+    private static final Map<String, BiConsumer<BookMetadata.BookMetadataBuilder, String>> CALIBRE_FIELD_MAPPINGS = Map.ofEntries(
+        Map.entry("#title", (builder, value) -> builder.title(value)),
+        Map.entry("#subtitle", (builder, value) -> builder.subtitle(value)),
+        Map.entry("#publisher", (builder, value) -> builder.publisher(value)),
+        Map.entry("#description", (builder, value) -> builder.description(value)),
+        Map.entry("#language", (builder, value) -> {
+            if (!value.isEmpty() && !"UND".equalsIgnoreCase(value)) {
+                builder.language(value);
+            }
+        }),
```
MightyHalfGinger: Should we really allow these duplicated tags? We already have them in the main attributes in the epub so it feels a bit weird to also have them as custom user_metadata... When calibre writes the metadata file it uses the proper fields for these at least.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+        Map.entry("#series_name", (builder, value) -> builder.seriesName(value)),
+        Map.entry("#series_number", (builder, value) -> safeParseFloat(value, builder::seriesNumber)),
```
MightyHalfGinger: calibre already has well-defined fields for this rather than custom columns so do we really need custom user_metadata support for this? I see it saving `calibre:series_index` and `calibre:series` on my system at least. If we want to keep this support for custom user_metadata as a source as well I think we should make the terminology line up.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+        Map.entry("#series_total", (builder, value) -> safeParseInt(value, builder::seriesTotal)),
+        Map.entry("#isbn_13", (builder, value) -> builder.isbn13(value)),
+        Map.entry("#isbn_10", (builder, value) -> builder.isbn10(value)),
+        Map.entry("#asin", (builder, value) -> builder.asin(value)),
+        Map.entry("#goodreads_id", (builder, value) -> builder.goodreadsId(value)),
+        Map.entry("#comicvine_id", (builder, value) -> builder.comicvineId(value)),
+        Map.entry("#hardcover_id", (builder, value) -> builder.hardcoverId(value)),
```
MightyHalfGinger: From what I can tell these ids are already stored as identifiers, and show up as such when importing a book into calbre. Do we really need custom user_metadata support for these?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -156,24 +238,18 @@ public BookMetadata extractMetadata(File epubFile) {
                                     } catch (NumberFormatException ignored) {
                                     }
                                 }
+                                
+                                if ("calibre:tags".equals(name)) {
```
MightyHalfGinger: Im not seeing this field ever used by calibre... Is there an example on how this can get set? Im fine with using this as where we save the separate "tags" for booklore, Im just a bit confused on where its coming from.

Another option would be a custom "booklore" field such as `booklore:tags` to make it clear that this is booklore's meaning of tags rather than calibre's. I see we have `calibre:user_metadata#extra_tags` storing the booklore tag data as well, so maybe thats what we should use instead for this...

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -410,4 +619,94 @@ private byte[] extractFileFromZip(File epubFile, String path) {
...
+                    if ("#extra_tags".equals(fieldName)) {
+                        extractAndSetUserMetadataSet(value, builder::tags);
+                        continue;
+                    }
+                    
+                    if ("#authors".equals(fieldName)) {
+                        extractAndSetUserMetadataSet(value, builder::authors);
+                        continue;
+                    }
+
```
MightyHalfGinger: Why would authors ever appear as an extra column when we have native support for it in the epub?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -198,31 +282,26 @@ public BookMetadata extractMetadata(File epubFile) {
                                     }
                                 }
                             }
-                            case "subject" -> categories.add(text);
+                            case "subject" -> tags.add(text);
```
MightyHalfGinger: This should continue to be saved to categories as before

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+        final BiConsumer<BookMetadata.BookMetadataBuilder, String> setter;
+
+        IdentifierMapping(String prefix, String fieldName, BiConsumer<BookMetadata.BookMetadataBuilder, String> setter) {
+            this.prefix = prefix;
+            this.fieldName = fieldName;
+            this.setter = setter;
+        }
+    }
+
+    private static final List<IdentifierMapping> IDENTIFIER_PREFIX_MAPPINGS = List.of(
```
MightyHalfGinger: Can we add the douban_id here rather than in the customer user_metadata?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -264,7 +343,7 @@ public BookMetadata extractMetadata(File epubFile) {
                     }
 
                     builderMeta.authors(creatorsByRole.get("aut"));
-                    builderMeta.categories(categories);
+                    builderMeta.tags(tags);
```
MightyHalfGinger, line 326: See my main comment, I think we can have both here still so long as we fix the attribute to append rather than overwrite

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+        Map.entry("#isbn_13", (builder, value) -> builder.isbn13(value)),
+        Map.entry("#isbn_10", (builder, value) -> builder.isbn10(value)),
+        Map.entry("#asin", (builder, value) -> builder.asin(value)),
+        Map.entry("#goodreads_id", (builder, value) -> builder.goodreadsId(value)),
+        Map.entry("#comicvine_id", (builder, value) -> builder.comicvineId(value)),
+        Map.entry("#hardcover_id", (builder, value) -> builder.hardcoverId(value)),
+        Map.entry("#hardcover_book_id", (builder, value) -> safeParseInt(value, builder::hardcoverBookId)),
+        Map.entry("#douban_id", (builder, value) -> builder.doubanId(value)),
+        Map.entry("#google_id", (builder, value) -> builder.googleId(value)),
+        Map.entry("#page_count", (builder, value) -> safeParseInt(value, builder::pageCount)),
```
MightyHalfGinger: Don't we handle `#page_count` manually in the metadata section below? We should either use this map or manual logic, not both, though I think I prefer this map style. If we decide to keep the map entry then we should probably also move `#pages` handling here as well.

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -410,4 +619,94 @@ private byte[] extractFileFromZip(File epubFile, String path) {
...
+                    if ("#moods".equals(fieldName)) {
+                        extractAndSetUserMetadataSet(value, builder::moods);
+                        continue;
+                    }
+                    
+                    if ("#extra_tags".equals(fieldName)) {
+                        extractAndSetUserMetadataSet(value, builder::tags);
+                        continue;
+                    }
+
```
MightyHalfGinger: Is there a reason we have these as explicit if blocks rather than entries in the `CALIBRE_FIELD_MAPPINGS`?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -141,6 +165,106 @@ public void writeMetadataToFile(File epubFile, BookMetadataEntity metadata, Stri
...
+            copyStringFieldToCalibre(helper::copyAsin, clear != null && clear.isAsin(), 
+                metadataElement, opfDoc, "#asin", "comments", hasChanges);
+            copyStringFieldToCalibre(helper::copyGoodreadsId, clear != null && clear.isGoodreadsId(), 
+                metadataElement, opfDoc, "#goodreads_id", "comments", hasChanges);
+            copyStringFieldToCalibre(helper::copyGoogleId, clear != null && clear.isGoogleId(), 
+                metadataElement, opfDoc, "#google_id", "comments", hasChanges);
+            copyStringFieldToCalibre(helper::copyHardcoverId, clear != null && clear.isHardcoverId(), 
+                metadataElement, opfDoc, "#hardcover_id", "comments", hasChanges);
+            copyStringFieldToCalibre(helper::copyComicvineId, clear != null && clear.isComicvineId(), 
+                metadataElement, opfDoc, "#comicvine_id", "comments", hasChanges);
```
MightyHalfGinger: See above, we already write these to the identifiers that calibre will pick up for us, do we really need to also put them in the user_metadata as well?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -141,6 +165,106 @@ public void writeMetadataToFile(File epubFile, BookMetadataEntity metadata, Stri
...
+            copyIntFieldToCalibre(helper::copyGoodreadsReviewCount, clear != null && clear.isGoodreadsReviewCount(), 
+                metadataElement, opfDoc, "#goodreads_review_count", hasChanges);
+
+            helper.copyHardcoverBookId(clear != null && clear.isHardcoverBookId(), val -> {
+                removeCalibreUserMetadataField(metadataElement, "#hardcover_book_id");
+                if (val != null && val > 0) {
+                    writeCalibreUserMetadataField(metadataElement, opfDoc, "#hardcover_book_id", String.valueOf(val), "comments", org.springframework.boot.configurationprocessor.json.JSONObject.NULL);
+                }
+                hasChanges[0] = true;
+            });
```
MightyHalfGinger, line 206: Shouldn't we save this as an integer with `copyIntFieldToCalibre` rather than a string?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -610,4 +786,232 @@ private void deleteDirectoryRecursively(Path dir) {
...
+            String newId = "title-main";
+            titleElement.setAttribute("id", newId);
+            return newId;
+        }
+        
+        return existingId;
+    }
+
+    private void createTitleRefinement(Element metadataElement, Document doc, String titleId, String titleType, String value) {
+        Element titleElement = doc.createElementNS("http://purl.org/dc/elements/1.1/", "title");
```
MightyHalfGinger: We should move the DC_NS constant out to the class level so we can just reference it here instead

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -610,4 +786,232 @@ private void deleteDirectoryRecursively(Path dir) {
...
+        }
+        
+        if (userMetaElement != null) {
+            metadataElement.removeChild(userMetaElement);
+            metadataElement.appendChild(userMetaElement);
+        }
+    }
+
+    private String ensureTitleHasId(Element metadataElement, Document doc, String dcNamespace) {
+        NodeList titleNodes = metadataElement.getElementsByTagNameNS(dcNamespace, "title");
```
MightyHalfGinger: Should we just hardcode the DC_NS here rather than taking it in as an argument? The title should always be defined in the DC_NS from the spec as I understand it.

booklore-api/src/test/java/com/adityachandel/booklore/service/metadata/writer/EpubMetadataWriterTest.java
```diff
@@ -126,8 +140,25 @@ void replaceCoverImageFromUpload_withUnicodeHref_handlesDecoding() throws IOExce
         }
     }
 
-    private File createEpubWithOpf(String opfContent, String filename) throws IOException {
-        File epubFile = tempDir.resolve(filename).toFile();
+    private byte[] createBasicEpub() throws IOException {
```
MightyHalfGinger: We should probably keep the opf version with custom data, then have the basic epub call use it internally

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -101,13 +182,14 @@ public BookMetadata extractMetadata(File epubFile) {
...
-                    Set<String> categories = new HashSet<>();
+                    Set<String> tags = new HashSet<>();
+                    Set<String> processedIdentifierFields = new HashSet<>();
 
                     boolean seriesFound = false;
                     boolean seriesIndexFound = false;
 
                     NodeList children = metadata.getChildNodes();
-
+
```
MightyHalfGinger: Why the whitespace change?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -140,9 +222,9 @@ public BookMetadata extractMetadata(File epubFile) {
                                 if ("title-type".equals(prop) && StringUtils.isNotBlank(refines)) {
                                     titleTypeById.put(refines.substring(1), content.toLowerCase());
                                 }
-
+
```
MightyHalfGinger: Why the whitespace change?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -1,5 +1,22 @@
...
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.sql.*;
+import java.util.*;
+import java.util.regex.Pattern;
+
```
MightyHalfGinger: Can we put these with the other imports rather than separately?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+    private static final List<IdentifierMapping> IDENTIFIER_PREFIX_MAPPINGS = List.of(
+        new IdentifierMapping("asin:", "asin", BookMetadata.BookMetadataBuilder::asin),
+        new IdentifierMapping("amazon:", "asin", BookMetadata.BookMetadataBuilder::asin),
+        new IdentifierMapping("mobi-asin:", "asin", BookMetadata.BookMetadataBuilder::asin),
+        new IdentifierMapping("goodreads:", "goodreadsId", BookMetadata.BookMetadataBuilder::goodreadsId),
+        new IdentifierMapping("google:", "googleId", BookMetadata.BookMetadataBuilder::googleId),
+        new IdentifierMapping("hardcover:", "hardcoverId", BookMetadata.BookMetadataBuilder::hardcoverId),
+        new IdentifierMapping("comicvine:", "comicvineId", BookMetadata.BookMetadataBuilder::comicvineId)
+    );
```
MightyHalfGinger, line 74: Can we add the douban_id here rather than in the custom user_metadata?

booklore-api/src/main/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -21,22 +39,85 @@
...
+
+    private static final Map<String, BiConsumer<BookMetadata.BookMetadataBuilder, String>> SCHEME_MAPPINGS = Map.of(
+        "GOODREADS", BookMetadata.BookMetadataBuilder::goodreadsId,
+        "COMICVINE", BookMetadata.BookMetadataBuilder::comicvineId,
+        "GOOGLE", BookMetadata.BookMetadataBuilder::googleId,
+        "AMAZON", BookMetadata.BookMetadataBuilder::asin,
+        "HARDCOVER", BookMetadata.BookMetadataBuilder::hardcoverId
+    );
```
MightyHalfGinger, line 82: Would need to DOUBAN here as well I think

jadehawk:
> I finally sat down and went through this huge change in total, and I'm a bit concerned about the complexity/overreliance on user_metadata from calibre here. The whole point is they are custom for a "user", and not something more standard that epubs usually work with. There are certain usecases like the reviews and such where it makes a lot of sense to store the data there, but in many of the other cases we have existing epub standards that we should be using instead so that epub readers can hopefully make use of the data.
> 
> As an example, the more I think about this the less I think we should really need to add such complex support for the genres... Looking at how calibre works, it just saves the `tags` section directly to the subject in the opf file without any more thought. I think we should just follow along here and leave it so that calibre's "tags" are equivalent to booklore's "genres". I know they can be completely arbitrary, but having tried to fetch metadata for various books, the genre data from goodreads/hardcover/etc. can also be all over the place rather than strict "genres". This way booklore's separate tags can be saved to some other attribute in the epub rather than the subjects (e.g. something like `booklore:tags`).
> 
> As a suggestion:
> 
> * booklore's genre/categories =writes to=> epub's `subject` =read as=> calibre's tags
> * booklore's tags =writes to=> epub's `booklore:tags` and `calibre:user_metadata#extra_tags` =read as=> calibre custom extra_tags data
> 
> That way the data is preserved on both sides and should survive imports in both directions while fitting in with most epub readers that will only support subjects.
> 
> > You raised a great question here. I did not use that function because of the way I was generating the "categories" value, and when using the builder, it puts the first value found when called and ignores the rest.
> 
> The issue seems to stem from us effectively "double calling" the builder functions in both the outer logic where we get it from the `calibre:tags` and the `subject` metadata, and then again from the logic inside `extractCalibreUserMetadata`. I made a quick mock setup locally where I called builderMeta.authors multiple times with different data each time and only the last set of values stuck as you stated.
> 
> Looking at [the lombok docs](https://projectlombok.org/features/Builder) it seems this is the default behaviour for fields, but it can be changed for collections like we have here by adding a `@Singular` tag to the attribute in `BookMetadata.java`. That seems to fix my mock setup, so perhaps that will fix what you're referring to here as well, if we want to have them be combined.

@MightyHalfGinger

Happy New Year!

Thank you for reviewing the code and providing feedback on what needed to be changed or refactored.

I agree with all your points and have adjusted my approach. I moved away from the overly complex #genre / #categories extraction from Calibre and simplified things by continuing to treat dc:subject the same way Booklore currently does—mapping it directly to Booklore’s Genres (categories).

Our focus is now strictly on importing Calibre’s “Ratings & Reviews” (Amazon, Goodreads, Hardcover) when those custom columns exist, along with the corresponding IDs found in the Identifiers section. We’re also bringing in #extra_tags, #moods, and #subtitle. This covers all metadata Booklore currently tracks.

We’re still maintaining redundancy of Booklore metadata inside the EPUB using booklore:* fields. This ensures that if the database fails or a user recreates their Booklore library, all UI‑visible metadata remains preserved inside the EPUB. Once a book becomes part of Booklore’s library, all Calibre tags are removed and replaced with booklore:* tags, resulting in a much cleaner EPUB structure.

I also cleaned up whitespace, restored constants, and brought back code that was accidentally removed.

Regarding Douban metadata: I left that section out because DoubanBookParser.java already collects and maps all Douban metadata to the appropriate Booklore fields. However, doubanId, doubanRating, and doubanReviewCount are discarded during database entry since they don’t exist in BookMetadataEntity. This might be an incomplete implementation or intentional design—hard to say. To persist Douban metadata, I’d need to add those fields to the database and complete the implementation.

About @singular: I initially used it as you suggested and it worked well. But when MetadataChangeDetectorTest was introduced in PR #1985 (Part of Release 1.16.1), it failed on all three fields where I used @singular (authors, tags, moods). Removing @singular resolved the issue and the tests passed. I also manually tested adding authors, saving, adding another author, saving again, and all authors were correctly written to the EPUB—not just the last one. This was after removing the complicated genre/category logic.

Not sure when you or other devs will have time to review the PR with the New Year, but until then—

Happy New Year, everyone!

SKlopp:
@MightyHalfGinger 
> As an example, the more I think about this the less I think we should really need to add such complex support for the genres... Looking at how calibre works, it just saves the `tags` section directly to the subject in the opf file without any more thought. I think we should just follow along here and leave it so that calibre's "tags" are equivalent to booklore's "genres". I know they can be completely arbitrary, but having tried to fetch metadata for various books, the genre data from goodreads/hardcover/etc. can also be all over the place rather than strict "genres". This way booklore's separate tags can be saved to some other attribute in the epub rather than the subjects (e.g. something like `booklore:tags`).


Some additional thoughts on the "tags"-question. From my POV {calibre tags} => {booklore genres} never made sense (tags are ... tags), but that's a design decision I don't want to question. 
So perhaps it would be a way out of this confusion to have a function to copy/move/swap genres <=> tags in booklore...

jadehawk:
@MightyHalfGinger @balazs-szucs @acx10 

Rebased and played Cat & Mouse adjusting code to fit with the updates and changes made to the metadata handling.
Code ready for review, hopefully it won't be 30+ commits again and I'll be back to working around metadata changes made by the real devs. If there is a plan to handle metadata differently, please close this PR, and I'll await whatever it is that you guys have planned.

Docker image available at:
```
docker pull jadehawk/jh-booklore:1.17.0-cali
```

balazs-szucs:
Hi,

Apologies for the unresponsiveness. From my side it looks good, though admittedly I mainly looked at the Java code quality rather than the actual functionality (I don't use Calibre, so I'm not quite certain what its peculiarities are). We've been a bit swamped lately, and large PRs have been ignored because of that (mine included OIDC one.). I don't plan to pick up any big PRs or projects for now (see this comment for context: https://github.com/booklore-app/booklore/pull/1734#issuecomment-3739762481); I think @acx10 will eventually have to get to this.

Sorry again; so there have been 3-4 bigger PRs affected by this (this included). Hopefully we can get this sorted shortly.

jadehawk:
> Hi,
> 
> Apologies for the unresponsiveness. From my side it looks good, though admittedly I mainly looked at the Java code quality rather than the actual functionality (I don't use Calibre, so I'm not quite certain what its peculiarities are). We've been a bit swamped lately, and large PRs have been ignored because of that (mine included OIDC one.). I don't plan to pick up any big PRs or projects for now (see this comment for context: [#1734 (comment)](https://github.com/booklore-app/booklore/pull/1734#issuecomment-3739762481)); I think @acx10 will eventually have to get to this.
> 
> Sorry again; so there have been 3-4 bigger PRs affected by this (this included). Hopefully we can get this sorted shortly.

@balazs-szucs @acx10 @MightyHalfGinger 

Figure as much, This PR started because Booklore was not reading all "dc:tags" from the EPUB (ASIN, GOOGLE, GOODREADS.....). So migrating from Calibre will mean not bringing over the metadata to Booklore.

PR added the choice to Calibre user to use Custom Columns to populate with the "Basic" Metadata Booklore currently accepts. (Title, Subtitle, Language, Publisher, Publish Date, Authors, Genres, Moods, Tags, Series Name, Series #, Series Total, Pages, ISBN-10, ISBN-13, ASIN, AZ Rating, AZ Reviews, GoogleID, GoodreadsID, GR Rating, GR Reviews, HardcoverID, HardcoverBookID, HC Rating, HC Reviews, ComicvineID --- And now newly added LC ID, LC Rating, RanobedbID, Ranobedb Rating)

All but the last 2 added metadata providers (never even heard of them) can be added by the use of custom columns in Calibre.

This was going to be it for the PR. But then I noticed Booklore does not save the metadata we just collected to the EPUB, as a Calibre user it was OK, my books already have metadata embedded, but as someone that may be just a sole Booklore user, no metadata persistence at all, DB corrupts user looses the basic metadata mentioned earlier and needs to fetch it again, not ideal.

So this PR evolved and also handles metadata persistence of ALL the metadata I just mentioned including the new ones into the EPUB.

So the Calibre angle is now just a "Bonus", the real benefit is the Metadata persistence into EPUB eBooks.

Originally It was only extracting metadata from EPUB during "BookDrop" only, with the 30+ changes I had to merge, now we can extract the same metadata from the EPUB by uploading to library or BookDrop.

Internally the EPUB gets cleaned of any Calibre tag, and we persist metadata by using booklore:* tags making it cleaner and less bloated than Calibre's method.

Perhaps a change of Title will be better for this PR?.
```
feat(metadata) Save basic metadata to EPUB file - Bonus use Calibre Custom Columns for importing metadata.
```
No matter what, Thank you for replying and just so you know all that you guys do is appreciated.

beedaddy:
It seems that this will break the detection of `#pagecount`, on which I currently rely. I can adjust the Calibre column name accordingly, that would be fine with me. However, we should document the custom Calibre columns supported by Booklore in the user manual. Then it is easy to see at a glance how these columns should be named.

jadehawk:
No need to adjust.  I never had Booklore detect pages for me. If you are saying, that Booklore Already detected #pagecount  and mapped it properly to the Pages metadata, then I'll adjust it on my end. But I did not find page count extraction during making this PR.

I'll check once I get home if #pagecount is currently working. 

Thx for the info.

jadehawk:
> It seems that this will break the detection of `#pagecount`, on which I currently rely. I can adjust the Calibre column name accordingly, that would be fine with me. However, we should document the custom Calibre columns supported by Booklore in the user manual. Then it is easy to see at a glance how these columns should be named.

@beedaddy 

Confirmed. The only Calibre Custom column I found was `#pagecount` which I had Assimilated(and forgot) and changed to `#page_count` to keep with the naming schema of the database, but did not pass my mind anyone else was using it as it was not documented anywhere. I have return the lookup name to `#pagecount` as not to mess with current Calibre/Booklore users flow.

Thank you for pointing this out. Booklore Test Docker Image Updated.

jadehawk:
@balazs-szucs @acx10 @MightyHalfGinger 

Good evening,

Rebased to v1.18.5, and conflicts resolved.

Docker test Image:
``` code
docker pull jadehawk/jh-booklore:1.18.5-metadata-saver
```

balazs-szucs:
Does this have the metadata as in a custom booklore namespace? Does it use the EPUB 3.X way of doing this e.g.,

```xml
<package ... prefix="booklore: http://booklore.org/metadata/1.0/">
  <metadata>
    <meta property="booklore:goodreads_rating">4.10</meta>
    <meta property="booklore:page_count">324</meta>
    <meta property="booklore:moods">Tense, Adventurous, Mysterious</meta>
  </metadata>
</package>
```

Or does this do same thing as Calibre? AFAICT, it does the latter. In that case, it's not really compliant with the EPUB standard. 

Did you do end to end testing? E.g., did you make library write, bunch metadata, delete the library and when you re-imported did the custom Metadata fields get populated?

jadehawk:
> Does this have the metadata as in a custom booklore namespace? Does it use the EPUB 3.X way of doing this e.g.,
> 
> ```
> <package ... prefix="booklore: http://booklore.org/metadata/1.0/">
>   <metadata>
>     <meta property="booklore:goodreads_rating">4.10</meta>
>     <meta property="booklore:page_count">324</meta>
>     <meta property="booklore:moods">Tense, Adventurous, Mysterious</meta>
>   </metadata>
> </package>
> ```
> 
> Or does this do same thing as Calibre? AFAICT, it does the latter. In that case, it's not really compliant with the EPUB standard.
> 
> Did you do end to end testing? E.g., did you make library write, bunch metadata, delete the library and when you re-imported did the custom Metadata fields get populated?

@balazs-szucs 

- c7d8193d: Add EPUB 3 namespace prefix for custom booklore metadata
- 7c500bfb: Move hardcoverBookId to standard URN identifier format
- Followed your commit on ISBN preservation (after Validation) Strips "-" Validates, preserves "-" when writing to EPUB.

Testing I performed is as follow on my custom docker image:

- DELETE MariaDB completely
- Loaded Booklore image.
- Added 10 books.
- FILLED All metadata fields with TEST data.
- Saved each book (Write metadata to EPUB is enabled in the settings)
- Downloaded all 10 books back to PC.
- DELETE MariaDB completely Again.
- Relaunched Booklore.
- Recreated Library.
- Imported to "LIBRARY" all 10 books.
- Re-scanned Library for books added.
- **All 10 books have ALL Fields POPULATED** (Successfully retrieve saved metadata)
- Repeated steps from "DELETE MariaDB completely" but imported books using "Bookdrop".
- **All 10 books have ALL Fields POPULATED** (Successfully retrieve saved metadata).

We are Saving:

- Standard metadata (title, authors, publisher, etc.)
- **Custom booklore metadata**: page_count, series_total, ratings, review counts, moods, tags
- **Provider IDs:** goodreadsId, hardcoverId, hardcoverBookId, googleId, comicvineId, lubimyczytacId, ranobedbId

As for EPUB 3 Compliance, the code:

- Uses prefix declaration: booklore: http://booklore.org/metadata/1.0/
- Uses <meta property="booklore:*"> format (not like Calibre)
- Uses URN identifiers: such as urn:hardcoverbook:{value}
- Metadata persists through delete/re-import cycle


_Just Awaiting the merge_.

ONLY ISSUE I FOUND, not something I did but was done by someone else way back.
The DATA type for "hardcover_book_id" does not match other Identifiers.

During DB Migration V74:
```code
-- Add numeric hardcover_book_id column to book_metadata table
-- This stores the numeric Hardcover book ID for API operations,
-- while the existing hardcover_id column stores the slug for URL linking.

ALTER TABLE book_metadata ADD COLUMN hardcover_book_id INTEGER;
```

If user attempts to save a hardcover book id larger than "2,147,483,647" (I know, is a huge number and ridiculous) You will get an error in the UI.

All Other like fields (hardcover_id,goodreads_id,google_id) use VARCHAR(100)

I figure I point this out. If Data type is ever changed to VARCHA(100). My code will also need to be updated for that particular field.

I had this in mind and update my code to fit:
`V100__Change_hardcover_book_id_to_varchar.sql`
```code
-- Change hardcover_book_id from INTEGER to VARCHAR(100) for consistency with other provider IDs
-- This prevents overflow issues with large book IDs (e.g., > 2,147,483,647)
ALTER TABLE book_metadata ALTER COLUMN hardcover_book_id TYPE VARCHAR(100);
```

As of right now. Code is clean IMHO and working as expected (Is the image I am running on my server regardless, I just don't want to keep updating just so I can be current with upstream)

balazs-szucs:
Hi, firstly, thanks for your patience and work. I really appreciate it.

@acx10, could you please merge this? It's had ample scrutiny as you can see, and it's a generally welcome improvement (and bugfix with hardcover).

I think the hardcover fix may also address some underlying bugs or issues we may have missed, so it's really great that you found that bug.

For the other formats, I can take over if you'd like (or if you were only interested in EPUBs). Essentially, the hope is that we can save all fields to all file formats. I have working code for PDFs (though I haven't made it to the testing stage with provider IDs, hence the Hardcover bit was a surprise for me too) and I'm currently working on Comics. I want to use similar structure, variable names, etc., as this one since this will be the "standard" for us going forward. This is to say, this is really helpful!

Thank you again!

balazs-szucs:
My hope generally, we can get to stage in next few releases where the write to file is stable enough, that people can delete their MariaDB, rescan and get back essentially the same library. It would be a really good place to be for the project, I think.

jadehawk:
> Hi, firstly, thanks for your patience and work. I really appreciate it.
> 
> @acx10, could you please merge this? It's had ample scrutiny as you can see, and it's a generally welcome improvement (and bugfix with hardcover).
> 
> I think the hardcover fix may also address some underlying bugs or issues we may have missed, so it's really great that you found that bug.
> 
> For the other formats, I can take over if you'd like (or if you were only interested in EPUBs). Essentially, the hope is that we can save all fields to all file formats. I have working code for PDFs (though I haven't made it to the testing stage with provider IDs, hence the Hardcover bit was a surprise for me too) and I'm currently working on Comics. I want to use similar structure, variable names, etc., as this one since this will be the "standard" for us going forward. This is to say, this is really helpful!
> 
> Thank you again!

@balazs-szucs 

The HardcoverBookId Issue I pointed out, is still active. I did not dare to update database. But we are at the very least aware of the issue.

If you approve I will update my code and add the DB migration I proposed (have it all written down on a separate file in the event I had to update code down the line) so that going forward, we won't encounter that possible "Edge Case" with the HardcoverBookid.

As for the other formats, I really only deal with EPUBS, and have no issues with you taking the lead on those.

Let me know on the DB migration. And I'll handle it fast.

balazs-szucs:
The database fix, would require migration, and that's generally a whole bigger headache :cry:, if you don't mind dealing with it, lets do it here. But its up to you :smile:

jadehawk:
> The database fix, would require migration, and that's generally a whole bigger headache 😢, if you don't mind dealing with it, lets do it here. But its up to you 😄

@balazs-szucs 

You had that correct.

UPDATE: migrate hardcover_book_id to VARCHAR and fix related issues

## Changes

### 1. Database Migration
- Changed hardcover_book_id column from INTEGER to VARCHAR(255)
- Supports alphanumeric book IDs from Hardcover API
- Updated HardcoverSyncService to handle String ↔ Integer conversion

### 2. Metadata Editor Fixes
- Fixed metadata change detection for provider-specific fields
- Added missing clearFlags entries: hardcoverBookId, lubimyczytacId, lubimyczytacRating
- Resolves issue where rating/review count fields wouldn't save independently

Affected fields now save properly independently:
- Amazon rating & review count
- Goodreads rating & review count
- Hardcover rating & review count
- Lubimyczytac rating
- Ranobedb rating

### 3. Web Reader Hardcover Sync (Major Fix)
Since I changed the data type for hardcover_book_id I had to update references in the HardCoverSyncService.java This turned out to reveal the Sync Service was not fully implemented I never used those tools within booklore till today during testing:
- Added Hardcover progress sync to ReadingProgressService
- Previously, HARDCOVER sync only worked for Kobo and KOReader devices
- Web browser reading progress now syncs to Hardcover.app

## Files Modified
- booklore-api/src/main/resources/db/migration/V107__Change_hardcover_book_id_to_varchar.sql
- booklore-api/src/main/java/.../service/hardcover/HardcoverSyncService.java
- booklore-api/src/main/java/.../service/progress/ReadingProgressService.java
- booklore-ui/src/app/features/metadata/.../metadata-editor.component.ts

## Breaking Changes
None - migration handles existing integer IDs gracefully

## Testing
- Hardcover sync tested with web reader
- Metadata editor field updates verified
- Database migration confirmed successful

As USUAL a docker image with current code updates is available for testing at:
```code
docker pull jadehawk/jh-booklore:1.18.5-metadatav4
```

acx10:
@balazs-szucs Thanks man! I’ll take your word for it.

jadehawk:
@balazs-szucs I know this was merged and supposed to have been added on v2.0, But a complete refactor was done on the EpubMetadataExtractor.java before v2.0 was released essentially removing all logic from this PR replaced by your logic that looks neat, yet none of the metadata that was previously successfully extracted from the epub (processed in Calibre) is ever reaching Booklore's UI. Not even the simplest things such as Identifiers (ASIN,MOBI-ASIN,GOODREADS,GOOGLE,HARDCOVER......) Non are extracted.

Technically nullifying this PR completely.

I see a json sidecar was implemented in the release, this sidecar does not extract the metadata from the epub completely either. it's limited to Page count and title series and description only?? No identifiers. Was this done deliverately or an oversight? I happily updated to v2.0.4 from my custom image, to take advantage of the new updates only to find out the last 200 ebooks uploaded from calibre to booklore, non-had metadata extracted.

included is a TEST ebook (Scrambled) used to TEST ALL FIELDS (Minus the new Age & Content rating.
[The Dark Cygnet Files - #04.00.zip](https://github.com/user-attachments/files/25620824/The.Dark.Cygnet.Files.-.04.00.zip)

balazs-szucs:
@jadehawk I dont believe my refactor did that, 100% oversight (from somewhere, probably not my PR). From what I can tell, my PR is not supposed to have adverse affect on Calibre things. But I can take a look remedy this, shorty.

Very sorry about that.

balazs-szucs:
One thing, that I think may have caused this, is we moved XML parsing to XXE safe methods, I can see how that may not have Calibre namespace among "safe" XML namespaces ergo it does not parse that in. (But this is a guess I am on my phone I will take closer)

I think you can probably understand, why we dont want to parse in untrusted XML data, but if that is the case, it should be easy fix. Otherwise, I didnt "nuke" your code, so I believe this will the case of few lines of code to fix. Obviously to DEBUG will be bit... Harder.

jadehawk:
Copy that. Thank you for looking into it. Hopefully is an easy fix.

balazs-szucs:
I also did NOT modify JUnit testing methods, ergo, if or when it failed, we were oblivious to it since, the testing suite still passed, indicating no apparent problems. :cry: Sadly, this is kinda how Software goes, unless, there is explicit tests for it, it really hard to gurantee no regression. JUnit tests _has_ to fail explicitly in order for us to easily catch these. Sorry again!

balazs-szucs:
So, I changed the variables in the BL when setting up metadata embedding for the other two formats to keep things consistent. However, I only updated the BL metadata mapper and missed Calibre's at the EPUB extractor.

Sorry!

balazs-szucs:
Also, many thanks for letting me know this, I made the PR for this, I would hope this will be merged/released soon, please, feel free to keep me in the loop or ping if anything else pops-up.

balazs-szucs:
Also forgot to say: 
image

this how it parsed in your sample epub. I'm assuming this is good? If you can take quick look if it missed anything it'd nice. Thanks!

Issue #1854, opened 2025-12-14
Some dialogs do not adjust to mobile screen sizes


Muppetteer:
# 🐛 Bug Report

## 📝 Description
During cleanup of dialogs on #1842 I noticed a few forms which are almost unusable if displayed on a mobile portrait orientation as they require too much width to display on a narrow screen.

## 🔄 Steps to Reproduce
Open any of these forms on a mobile sized screen:
1. Library > Custom Fetch Metadata (Metadata Refresh Options)
2. Book > Custom Fetch Metadata (Metadata Refresh Options)
3. Book > Metadata fetch & copy / review
4. Organize Book Files
5. Bookdrop

**Result:**
The form design is such that controls do not wrap and the dialog is unable to display them adequately or without some non-obvious horizontal scrolling required.

## ✅ Expected Behavior
Display the form with controls wrapped / shrunk to fit.

## 💻 Environment
- **Booklore Version:** v1.14.1

PR #1911, opened 2025-12-16
fix: Bookdrop UI mobile support

Muppetteer:
# 🚀 Pull Request

## 📝 Description
Fixed the Bookdrop UI so that it handles better on mobile devices. Relates to #1854

## 🛠️ Changes Implemented

In desktop mode:
- Removed minimum width to allow mobile optimisation.
- Added cover image placeholder so book titles line up vertically when one is missing.
- Reduced size of library and subpath selection controls and aligned all vertically.

In mobile mode:
- Controls and text wrap and align to fit in portrait orientation.
- Hid some unnecessary items for better space allocation (eg tiny cover icon previews).
- Mobile specific footer buttons without text.

Both modes:
- Removed green metadata fetch bullet indicator, replaced with coloured/outlined right side chevron button (it was too hard to fit it on the mobile view).
- Display all fetched fields as disabled instead of just the tag chip fields.
- Only show paginator if there is more than one page.

## 🧪 Testing Strategy
In browser dev tools switching back and forth between desktop and mobile modes.

New file selection:
bookdrop-mobile-new

New file metadata picker:
bookdrop-meta-mobile-new

New desktop mode:
bookdrop-full-new

### DIFF
--- a/booklore-ui/src/app/features/bookdrop/component/bookdrop-file-review/bookdrop-file-review.component.html
+++ b/booklore-ui/src/app/features/bookdrop/component/bookdrop-file-review/bookdrop-file-review.component.html
@@ -1,388 +1,389 @@
-<div class="container">
-  <div class="main-card">
-    <div class="header">
-      <div class="header-content">
-        <h2>
-          Review Bookdrop Files
-          <a href="https://booklore-app.github.io/booklore-docs/docs/bookdrop" target="_blank" rel="noopener noreferrer">
-            <i
-              class="pi pi-external-link external-link-icon"
-              pTooltip="View documentation"
-              tooltipPosition="top"></i>
-          </a>
-        </h2>
-        <p>
-          These files were uploaded to the
-          <strong>Bookdrop Folder</strong>.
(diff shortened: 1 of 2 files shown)


acx10:
@Muppetteer Everything looks great except two UI glitches. Here's the video:

https://github.com/user-attachments/assets/5fe0c708-3cc6-4cb4-9089-c0764670e0c7

Thanks!

Muppetteer:
Both of those are existing problems with the current UI too. 

I have worked out the 2nd one. It is due to a default style rule applied somewhere in the build chain that targets all divs with a class "container". Will post a fix after I work out the first issue.

The 1st one is the primeng select component. It should auto align to the right if the drop down would appear off screen.

Muppetteer:
Actually I can't reproduce the first one. It looks fine for me on both Firefox and Chrome. Maybe a browser specific issue or bug with the primeng component?

looks-ok

Muppetteer:
I've fixed the 2nd issue and could not reproduce the 1st. However a long directory path could overflow on a mobile width screen so now the options wrap if too long which is a small improvement (mobile only).

PR #1912, opened 2025-12-16
feat: change image to a non root user

chkpwd:
Small changes that can go along way for users running Booklore inside restricted environments like Kubernetes. 

The container now runs as a non-root user by default and no longer relies on `su-exec`. All required `nginx` directories are created at build time, avoiding privileged filesystem changes during startup. The entrypoint was simplified to start nginx and the JVM directly using exec insted.

From su-exec:
> Notice that su-exec depends on being run by the root user, non-root users do not have permission to change uid/gid.

The nginx configuration was also simplified: runtime templating and file move operations were removed because they required root access, and port handling was clarified to avoid unnecessary constraints on host/container port mapping.

Documentation and examples were updated to remove deprecated USER_ID / GROUP_ID configuration and to better reflect the new, simpler runtime model.

As for the port handling, as far as I can tell, I don't see a reason why the angular app or nginx need both the host and container port to match. I've tested deployment of the app and its features like: opds, koreader sync and few others with success on separate ports.

image
 
To test, I deployed Booklore, ran normal operations like uploading, ingesting, and modifying books and their respective metadata with no errors in the console or container logs.

In this PR, I've added changes to the readme as well to remove references to previous *USER_ID and *GROUP_ID.

```bash
~ $ id
uid=65534(nobody) gid=65534(nobody) groups=65534(nobody)
❯ docker run --entrypoint /bin/sh -it -u 65534:65534 --rm booklore-nobody-nginx
~ $ ls -la /var/lib
total 0
drwxr-xr-x 1 root  root  10 Dec 28 16:10 .
drwxr-xr-x 1 root  root  18 Dec 28 16:10 ..
drwxr-xr-x 1 root  root   0 Oct  8 09:29 misc
drwxrwxrwx 1 nginx nginx 42 Dec 28 16:10 nginx
~ $ ls -la /var/log
total 0
drwxr-xr-x 1 root  root  10 Dec 28 16:10 .
drwxr-xr-x 1 root  root  18 Dec 28 16:10 ..
drwxrwxrwx 1 nginx nginx  0 Dec 28 16:10 nginx
~ $ ls -la /run/
total 0
drwxr-xr-x 1 root  root  10 Dec 28 16:10 .
drwxr-xr-x 1 root  root  12 Dec 28 17:07 ..
drwxr-xr-x 1 root  root   0 Oct  8 09:29 lock
drwxrwxrwx 1 nginx nginx  0 Dec 28 16:10 nginx
~ $
```

Before modifying installation instructions for `booklore-docs`, is there anything else that would be needed to change?


WorldTeacher:
Hi, I just saw this PR and had a small question:

what would happen if this is used on systems that prefer user:group allocations?
In my case, I am using unraid which by default assigns data the USER_ID of 99 and GROUP_ID of 100

I haven't used kubernetes yet, so I am not sure how stuff is managed there, but in other systems you can change this to the non-root user by setting the PUID and PGID, does that not work for kubernetes?

chkpwd:
> but in other systems you can change this to the non-root user by setting the PUID and PGID, does that not work for kubernetes?

PUID / PGID are not a container standard. 

They’re a convention popularized by LinuxServer.io images, where the container starts as root and switches to the requested UID/GID at runtime.

Depending on the Kubernetes environment, containers are often not allowed to start as root at all, which means they cannot change UID/GID at runtime. This results in apps like su-exec, gosu, or similar apps, simply not working. (They require root access to switch or elevate to other users)

Kubernetes handles this externally via pod security settings (runAsUser, runAsGroup, fsGroup) and expects the image itself to already function as a non-root user. (Not always the case, but you get the point)

This PR aligns BookLore with that model by removing runtime user switching and running non-root from the start. For Unraid and similar systems, permission alignment is still possible, but it needs to be handled at the host or Docker level rather than inside the container.

The main culprit to this escalation to root is nginx. Since we cannot change the final image to use an unprivileged nginx image, we can instead address the problem by creating the necessary subdirectories needed for nginx using the correct permissions.

@WorldTeacher

WorldTeacher:
okay, thanks for the clarification. Let's see what @acx10 thinks about this

I'll be looking into kubernetes after my vacation, I may have to do some changes to my services as well.

chkpwd:
No worries - if you need a quick intro into the world of k8s, give me a ping.

nginx.conf
```diff
@@ -14,8 +14,8 @@ http {
     large_client_header_buffers 8 32k;
 
     server {
-        listen ${BOOKLORE_PORT};
-        listen [::]:${BOOKLORE_PORT};
+        listen 6060;
+        listen [::]:6060;
```
acx10, line 18: Removing BOOKLORE_PORT will break existing user setups, as some users may be running Booklore on a port other than 6060. A couple of months ago I had to explicitly add BOOKLORE_PORT for backward compatibility for the same reason when I made changes to nginx stuffs.

I agree this isn’t the cleanest solution, but at this point we need to preserve compatibility for existing installations and deal with the cleanup more deliberately later rather than breaking users unexpectedly.

So we must keep the BOOKLORE_PORT.

You can reproduce this yourself fairly easily:
- Run the current (v1.16.5) Booklore image with a non-default port mapping (for example 6061:6061, which some users may have).
- Then update to this PR's image (I built one for you [pr-2165-51759c31](https://hub.docker.com/layers/booklore/booklore/pr-2165-51759c31/images/sha256-b86ef59cdab624c7852c49e4e01b2a51c93db03773b3b7dd635a5e04b75ae807)).

After the update, the UI will no longer come up unless the port is changed back to 6060:6060. This effectively breaks existing setups that are not using the default port, so removing the configurable port would be a breaking change for current users.

Dockerfile
```diff
@@ -51,14 +50,24 @@ LABEL org.opencontainers.image.title="BookLore" \
...
 
-RUN apk update && apk add nginx gettext su-exec
+RUN apk add --no-cache nginx gettext \
+    && rm -rf /var/cache/apk/* \
+    && install -d -m 777 \
+        /run/nginx \
+        /var/lib/nginx \
+        /var/lib/nginx/tmp \
+        /var/lib/nginx/logs \
+        /var/log/nginx
```
acx10, line 60: This looks good!

Dockerfile
```diff
@@ -51,14 +50,24 @@ LABEL org.opencontainers.image.title="BookLore" \
...
 
-COPY ./nginx.conf /etc/nginx/nginx.conf
+COPY nginx.conf /etc/nginx/nginx.conf
 COPY --from=angular-build /angular-app/dist/booklore/browser /usr/share/nginx/html
 COPY --from=springboot-build /springboot-app/build/libs/booklore-api-0.0.1-SNAPSHOT.jar /app/app.jar
 COPY start.sh /start.sh
+
 RUN chmod +x /start.sh
 
+USER 65534:65534
```
acx10, line 69: Changing the container to run as USER 65534:65534 will break existing users’ ability to write to the library folder. Any file uploads or write operations will fail with errors like java.nio.file.AccessDeniedException, because the container no longer has permission to access host-mounted folders. 

We need to keep existing users in mind, many rely on the current UID setup for file access. A similar change in a previous PR caused issues, so we should avoid breaking current setups.

start.sh
```diff
@@ -18,4 +7,4 @@ mv "$TMP_CONF" /etc/nginx/nginx.conf
 nginx -g 'daemon off;' &
 
 # Start Spring Boot in foreground
-su-exec ${USER_ID:-0}:${GROUP_ID:-0} java -jar /app/app.jar
+exec java -jar /app/app.jar
```
acx10, line 10: To maintain backwards compatibility, we can conditionally use su-exec only when USER_ID and GROUP_ID are set, and otherwise run the container as the default (root) user, something like:

```
if [ -n "$USER_ID" ] && [ -n "$GROUP_ID" ]; then
    echo "Running as UID:$USER_ID GID:$GROUP_ID"
    exec su-exec "$USER_ID:$GROUP_ID" java -jar /app/app.jar
else
    echo "Running as root"
    exec java -jar /app/app.jar
fi
```

nginx.conf
rumblpak: The issue for k8s is not that the port exits, it's the mv tmp file to overwrite which unless the pod is privileged will be blocked. Have you considered just spawning nginx pointing at the tmp nginx file after making changes to it? That should fix both cases and the mv to overwrite /etc/nginx/nginx.conf can just be removed.

basically, instead of:
```
nginx -g 'daemon off;' &
```

```
nginx -c ${TMP_CONF} -g 'daemon off;' &
```

Dockerfile
shadofall: removing k8s for a moment, but the principle should be the same, the goal should be that the container starts with what ever is specified as a user , in compose terms this would be user: statement. which could be set to be the USER_ID and GROUP_ID ENV, i agree minting the backwards compatibility so that users who don't know wont get hit. so just need to make sure nothing needs root (by default docker will still run it as root regardless unless the user parameter is passed, )

and then ensure everything is flexible enough to use either the running user which, or the ENV's if specified.

there are other applications out there that will accept user:  or PUID/GUID configs, including both being set at the same time,, as long as someone doesnt do something like set user to 2000  and a uid to 2021 its fine., the tricky part is getting to not needing root to begin with.  which gets complicated when dealing with ngnix, but is a good goal to get twards.

PR #1987, opened 2025-12-25
Fix(metadata): Improve handling of non-author creator roles

MightyHalfGinger:
# 🚀 Pull Request

## 📝 Description
Right now booklore only correctly supports the author creator role in epubs, but assumes that all creator tags in epubs are authors rather than properly filtering based on the role. This makes epubs from some publishers that tag things like illustrators, translators, and even editors with creator tags end up getting incorrectly imported into booklore, which then confuses the metadata search when it ends up trying to use the editor for the author name when searching...

This could be solved one of two ways that I can think of: only pulling in the author role (and any un-role-d creators), or properly handling all of the roles that opf supports. For this PR I'm going with the former partially because its much simpler, and partially because I'm not quite sure how non-author roles should actually be tracked in the DB/UI anyway. Having the other roles in booklore would be amazing for sure, but I just want to avoid clobbering the author for now. I've designed the changes to at least make it easy to add in handling for other roles in the future.

## 🛠️ Changes Implemented
- Adds handling to the meta tags to track creator refinements by the refined id
- Adds in handling for opf:role attributes for creators
- Adds in handling to tracking creators by ID if necessary for refinements
- Improves the author writing logic to only remove creator tags for authors before writing the new values

## 🧪 Testing Strategy
I verified this with some epubs I have with extra creators tagged for illustrator, translator and editor. I confirmed that both the imported author was now correct, and that when the EPUB metadata was changed and written back to the file it only overwrote the actual author entries and not those for the illustrator, etc.

I also added testcases for both the extractor and writer that verify the expected parsing behaviour for non-author creators.

- [ ] Flyway migration versioning follows correct sequence _(if database schema modified)_

## 💬 Additional Context
Decided to make this change since I have dozens of epubs from a publisher that uses these sorts of creator tags, and it was making importing them and searching for metadata very annoying. I know I'll end up getting more from this publisher as well so I'd up running into this over and over in the future.

### DIFF
--- a/booklore-api/src/test/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractorTest.java
+++ b/booklore-api/src/test/java/com/adityachandel/booklore/service/metadata/extractor/EpubMetadataExtractorTest.java
@@ -233,6 +233,21 @@ class EpubMetadataExtractorTest {
         }
 
         @Test
+        @DisplayName("Should not extract non-authors from EPUB metadata")
+        void extractMetadata_withExtraCreators_returnsOnlyAuthors() throws IOException {
+            File epubFile = createEpubWithExtraCreators(DEFAULT_TITLE, DEFAULT_AUTHOR, "Jane Smith", "Alice", "Bob");
+
+            BookMetadata result = extractor.extractMetadata(epubFile);
+
+            assertNotNull(result);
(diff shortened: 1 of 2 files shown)


MightyHalfGinger:
Ugh, I hadn't meant to file the PR yet, will re-file this one again alter with the proper data in it.

PR #2067, opened 2025-12-31
fix(shelves): allow users without Manage Library permissions to edit and delete shelves

xcashy:
# 🚀 Pull Request

## 📝 Description

Previously, only users who had the Manage Library permissions were able to edit or delete shelves. This change updates the permission checks to allow any users to edit or remove shelves.

## 🛠️ Changes Implemented

- Adjust UI permission checks in book-browser.component.ts

- [x] Flyway migration versioning follows correct sequence _(if database schema modified)_

### DIFF
--- a/booklore-ui/src/app/features/book/components/book-browser/book-browser.component.html
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-browser.component.html
@@ -29,7 +29,7 @@
 
           @if (userService.userState$ | async; as userState) {
             @if (entityType !== EntityType.ALL_BOOKS && entityType !== EntityType.UNSHELVED &&
-            (userState.user!.permissions.admin || userState.user!.permissions.canManageLibrary)) {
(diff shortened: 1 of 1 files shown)


acx10:
If I’m not mistaken, users can already delete their own shelves and magic shelves, right?

xcashy:
> If I’m not mistaken, users can already delete their own shelves and magic shelves, right?

Yes, my mistake. Users can already delete shelves and magic shelves through the sidebar without this change. However, deleting them via the top bar requires the Manage Library permission.

acx10:
The fix makes sense for the case you described, but the condition as written shows the edit/delete menu for shelves and magic shelves to *all* users, not just the owner. If someone is viewing another user's shelf, they'd see options they can't actually use (or worse, can use but shouldn't). Could you scope it to check shelf ownership instead of just entity type?

PR #2068, opened 2025-12-31
feat: Add support for folder as series

### LINKED ISSUE #1461: New Import Mode: Each Folder is a Series.
**What problem or limitation are you encountering?**  
Magazines typically don't have medadata associated with them, or it's just left over metadata from the creation or scanning process leading to wildly inconsistent results when the scanner tries to read Title or series from the tagged metadata, and a massive mess of files if you're importing very long running magazines (National Geographic, Time, Rolling Stone, etc all have thousands of issues each)

**What solution or improvement do you propose?**  
In addition to "Each Folder is a book" or "Each File is a book", another import option like "Each Folder is a series". Even if it grabs the wrong series name, it would still drastically improve the organization of the library and make it easier for the user to edit the series media data as necessary. Would also be helpful for untagged comics, since many comic users will have series in folders.

CharlieQLe:
## Description
Adds support for using folders as the series for any books in its path. The primary use-case, for me at least, is to separate multiple comic series of the same name despite being published in different years.

Fixes #1461 

**Edit**: It appears that restarting the application resets the book series to whatever it has in its metadata at the moment

## Changes
- **Backend**: Adds a new library processor based off the existing file as book processor
- **Frontend**: Adds a new option for the library type dropdown

## Testing
Haven't tested extensively, nor have I written any tests. I did try using it with my own comic library following the naming convention recommended by Kavita.

## Visual Changes
image

## Checklist
- [ ] Automated unit/integration tests added/updated to cover changes
- [ ] All tests pass locally (`./gradlew test` for backend)
- [ ] Manual testing completed in local development environment

### DIFF
--- a/booklore-api/src/main/java/com/adityachandel/booklore/service/upload/FileUploadService.java
+++ b/booklore-api/src/main/java/com/adityachandel/booklore/service/upload/FileUploadService.java
@@ -89,28 +89,29 @@ public class FileUploadService {
     public AdditionalFile uploadAdditionalFile(Long bookId, MultipartFile file, AdditionalFileType additionalFileType, String description) {
         final BookEntity book = findBookById(bookId);
         final String originalFileName = getValidatedFileName(file);
+        final String sanitizedFileName = PathPatternResolver.truncateFilenameWithExtension(originalFileName);
 
         Path tempPath = null;
         try {
-            tempPath = createTempFile(UPLOAD_TEMP_PREFIX, originalFileName);
+            tempPath = createTempFile(UPLOAD_TEMP_PREFIX, sanitizedFileName);
             file.transferTo(tempPath);
 
             final String fileHash = FileFingerprint.generateHash(tempPath);
             validateAlternativeFormatDuplicate(additionalFileType, fileHash);
 
-            final Path finalPath = buildAdditionalFilePath(book, originalFileName);
+            final Path finalPath = buildAdditionalFilePath(book, sanitizedFileName);
             validateFinalPath(finalPath);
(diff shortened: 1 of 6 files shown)


acx10:
Hi @CharlieQLe ,

I have a few questions regarding this implementation. From what I understand, this PR currently treats folders as series during library creation and scanning, which makes sense.

However, I’m curious about scenarios where users leverage features that modify file paths or names, such as “Move via Pattern” or “Auto-move on metadata write”. In these cases, the folder structure could change after the initial scan, which might break the “series as folder” logic or cause inconsistencies with how books are grouped.

Have you considered how these features would interact with the new folder-as-series flow, or if any safeguards might be needed to preserve the intended series structure?

CharlieQLe:
> Hi @CharlieQLe ,
> 
> I have a few questions regarding this implementation. From what I understand, this PR currently treats folders as series during library creation and scanning, which makes sense.
> 
> However, I’m curious about scenarios where users leverage features that modify file paths or names, such as “Move via Pattern” or “Auto-move on metadata write”. In these cases, the folder structure could change after the initial scan, which might break the “series as folder” logic or cause inconsistencies with how books are grouped.
> 
> Have you considered how these features would interact with the new folder-as-series flow, or if any safeguards might be needed to preserve the intended series structure?

Looking back on it, this PR is more like a proof of concept than an actual feature implementation. I don't personally use the automatic move features as I use Booklore as a Kobosync server, so that wasn't initially considered in this PR.

This may have to work at the metadata extraction level like #2072, but that opens up its own can of worms if moving stuff between libraries. Maybe someone could pick up the torch on this one since the aforementioned PR is what this PR was going to do for me.

PR #2501, opened 2026-01-26
fix(ranobedb): fix search failing for titles with special characters

Ari-03:
**Summary**
- URL-encode search terms to handle special characters (e.g., `&` in "Wolf & Parchment")
- Remove duplicate `query` param (RanobeDB API only uses `q`)
- Remove English language filter (`rl=en`) that excluded Japanese-only books
**Testing**
Verified searches now work for:
- "Wolf & Parchment" (special char issue)
- "Maoyuu Maou Yuusha" (no English release)


acx10 (commented):
@Ari-03 Could you please add unit tests to verify that everything is functioning correctly?

PR #2597, opened 2026-02-03
feat(api, build): Upgrade to Java 25, Gradle 9, and Spring Boot 4, Jackson 3

balazs-szucs:
### 📝 Description

Since early December, as far as I can tell they fixed upstream the problem that caused Kobo problems. According to: https://hibernate.atlassian.net/browse/HHH-19890

The fix was released few days ago, at 23rd Jan.

### 🛠️ Changes Implemented

- 

### 📸 Visual Changes _(if applicable)_

---

### **Please Read - This Checklist is Mandatory**

> **Important Notice:** We've experienced several production bugs recently due to incomplete pre-submission checks. To maintain code quality and prevent issues from reaching production, we're enforcing stricter adherence to this checklist.
>
> **All checkboxes below must be completed before requesting review.** PRs that haven't completed these requirements will be sent back for completion.

#### **Mandatory Requirements** _(please check ALL boxes)_:

- [X] **Code adheres to project style guidelines and conventions**
- [X] **Branch synchronized with latest `develop` branch** _(please resolve any merge conflicts)_
- [X] **🚨 CRITICAL: Automated unit tests added/updated to cover changes** _(MANDATORY for ALL Spring Boot backend and Angular frontend changes - this is non-negotiable)_
- [X] **🚨 CRITICAL: All tests pass locally** _(run `./gradlew test` for Spring Boot backend, and `ng test` for Angular frontend - NO EXCEPTIONS)_
- [X] **🚨 CRITICAL: Manual testing completed in local development environment** _(verify your changes work AND no existing functionality is broken - test related features thoroughly)_
- [ ] **Flyway migration versioning follows correct sequence** _(if database schema was modified)_
- [ ] **Documentation PR submitted to [booklore-docs](https://github.com/booklore-app/booklore-docs)** _(required for features or enhancements that introduce user-facing or visual changes)_

#### **Why This Matters:**

Recent production incidents have been traced back to:

- **Incomplete testing coverage (especially backend)**
- Merge conflicts not resolved before merge
- Missing documentation for new features

**Backend changes without tests will not be accepted.** By completing this checklist thoroughly, you're helping maintain the quality and stability of Booklore for all users.

**Note to Reviewers:** Please verify the checklist is complete before beginning your review. If items are unchecked, kindly ask the contributor to complete them first.

---

### DIFF
--- a/booklore-api/src/main/java/org/booklore/service/library/LibraryService.java
+++ b/booklore-api/src/main/java/org/booklore/service/library/LibraryService.java
@@ -260,9 +260,8 @@ public class LibraryService {
 
     @Transactional
     public void deleteLibrary(long id) {
-        if (!libraryRepository.existsById(id)) {
-            throw ApiError.LIBRARY_NOT_FOUND.createException(id);
-        }
+        LibraryEntity library = libraryRepository.findById(id)
+                .orElseThrow(() -> ApiError.LIBRARY_NOT_FOUND.createException(id));
         monitoringService.unregisterLibrary(id);
         Set<Long> bookIds = library.getBookEntities().stream().map(BookEntity::getId).collect(Collectors.toSet());
         fileService.deleteBookCovers(bookIds);


balazs-szucs:
Hi @acx10,

Sorry for the, if you have the time can you test this with Kobo? (I don't a Kobo device)

As far as I can see the issue you linked last time that led to us reverting back to Java 21, has been resolved as per my PR description, so I assume if there isn't another popping up, then there is good chance this is good. I've been testing past few hours, for me so far so good, but sadly there only so much I can test without Kobo. 


Thanks!

balazs-szucs:
Hi @acx10,

We are actively testing this. You can see our convo at the dev-channel. We are optimistic that this works well! :smile:

acx10:
@balazs-szucs Thanks! I will give it a go tonight.

balazs-szucs:
Thanks! I have to push 1 commit to fix, a minor problem with auto-fetch. But that should done in less 10 mins. And i think its ready afterwards :smile:

balazs-szucs:
Apologies, this ready should be good, I just merged something, somewhere that caused me some headache

booklore-api/src/main/java/org/booklore/service/appsettings/AppSettingService.java
```diff
@@ -8,19 +9,20 @@
...
+import tools.jackson.core.JacksonException;
+import tools.jackson.core.type.TypeReference;
 
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 
 @Service
+@DependsOnDatabaseInitialization
```
acx10, line 25: Is this something required by Spring 4?

booklore-api/src/main/java/org/booklore/service/kobo/KoboServerProxy.java
```diff
@@ -128,11 +128,11 @@ private ResponseEntity<JsonNode> executeProxyRequest(HttpServletRequest request,
             HttpHeaders responseHeaders = new HttpHeaders();
             response.headers().map().forEach((key, values) -> {
                 if (isKoboHeader(key)) {
-                    responseHeaders.put(key, values);
+                    responseHeaders.addAll(key, values);
```
acx10, line 131: Are these actually equivalent? It doesn’t seem so to me.

booklore-api/src/main/java/org/booklore/service/kobo/KoboServerProxy.java
```diff
@@ -128,11 +128,11 @@ private ResponseEntity<JsonNode> executeProxyRequest(HttpServletRequest request,
...
                 }
             });
 
-            if (responseHeaders.containsKey(KoboHeaders.X_KOBO_SYNCTOKEN) && includeSyncToken && syncToken != null) {
+            if (responseHeaders.getFirst(KoboHeaders.X_KOBO_SYNCTOKEN) != null && includeSyncToken && syncToken != null) {
```
acx10, line 135: Are these actually equivalent? It doesn’t seem so to me.

Dockerfile
```diff
@@ -49,7 +49,9 @@ LABEL org.opencontainers.image.title="BookLore" \
       org.opencontainers.image.version=$APP_VERSION \
       org.opencontainers.image.revision=$APP_REVISION \
       org.opencontainers.image.licenses="GPL-3.0" \
-      org.opencontainers.image.base.name="docker.io/library/eclipse-temurin:21.0.9_10-jre-alpine"
+      org.opencontainers.image.base.name="docker.io/library/eclipse-temurin:25-jre-alpine"
+
+ENV JAVA_TOOL_OPTIONS="-XX:+UseG1GC -XX:MaxGCPauseMillis=200 -XX:+UseStringDeduplication -XX:+UseContainerSupport -XX:MaxRAMPercentage=75.0"
```
acx10, line 54: What’s the purpose of this JAVA_TOOL_OPTIONS configuration? Just trying to understand why we’re explicitly setting these GC and memory flags here.

booklore-ui/src/app/features/settings/task-management/task.service.ts
```diff
@@ -35,6 +35,7 @@ export interface LibraryRescanOptions {
 
 export interface TaskCreateRequest {
   taskType: TaskType;
+  triggeredByCron?: boolean;
```
acx10, line 38: What is this for?

balazs-szucs: Not sure which update triggered this, I would imagine Jackson 3, but it is strict about having to provide all fields explicitly, otherwise it will fail to parse. So, before this frontend "incomplete" request to the backend, and it'll error 500 with something like "Failed to parse primitive fields boolean" or something.

Dockerfile
balazs-szucs: These are standard Java flags for containerized usage. I was surprised that we hadn't used them before. The documentation on these is quite extensive; I was mainly exposed to this practice via Azure: https://learn.microsoft.com/en-us/azure/container-apps/java-memory-fit?tabs=create, which recommends it. However, it's also used in other OSS Java projects that are mainly distributed via Docker.

booklore-ui/src/app/features/settings/task-management/task.service.ts
acx10: Makes sense. Do you think this could be problematic in other areas though? If Jackson is now stricter about missing fields, we might run into similar issues in parts of the code we haven’t thoroughly tested yet.

booklore-api/src/main/java/org/booklore/service/appsettings/AppSettingService.java
balazs-szucs: Indirectly. It is more of Java 25 thing, that backend during testing (for me),  initialized too quickly leading to problems flyway problems.

TLDR; not required, it does not hurt, but can save some headaches in environment with beefy CPUs that can start the server too quickly.

There may be better solutions, though.

booklore-ui/src/app/features/settings/task-management/task.service.ts
balazs-szucs: For context this is for the "auto-fetch" on the book-browser view.

Dockerfile
acx10: Thanks for the explanation! We probably should have added these flags much earlier, they could explain some of the memory bloat issues people have been seeing.

booklore-api/src/main/java/org/booklore/service/kobo/KoboServerProxy.java
balazs-szucs: Yes, they are a bit different. What I did is a better in java 25, because `addAll` method ensures that all values are added to the internal storage of the `HttpHeaders` object. Most importantly, `HttpHeaders` will create a new, modifiable list if one doesn't exist yet for that key. However, with put there is no such guarantee, it would in that case fail. Essentially, it's less prone to error, but functionality wise it the same. (the all suggest adding all fields from the perspective of responseHeaders)

booklore-api/src/main/java/org/booklore/service/kobo/KoboServerProxy.java
acx10: I think containsKey(...) checks whether the header exists at all, while getFirst(...) != null specifically checks if the first value is non-null, so the two aren’t strictly equivalent. Right?

booklore-api/src/main/java/org/booklore/service/kobo/KoboServerProxy.java
acx10: Got it.

booklore-api/src/main/java/org/booklore/service/kobo/KoboServerProxy.java
balazs-szucs: Similar to above. getFirst is safer because it checks the actual values, returning null (and thus failing the != null check) when the header list is empty. containsKey, however, returns true as long as the key exists, even if the value list is empty, so it may lead to false positives. The two go hand-in-hand basically (meaning this, and change you commented above)

acx10: Looks good, I’ll go ahead and merge the PR once the merge conflicts are resolved. There are just a couple small ones to fix. Thanks!

booklore-ui/src/app/features/settings/task-management/task.service.ts
balazs-szucs: > Makes sense. Do you think this could be problematic in other areas though? If Jackson is now stricter about missing fields, we might run into similar issues in parts of the code we haven’t thoroughly tested yet.

I haven't encountered this elsewhere, but yes it is a possibility. It does not crash the app, just throws an error. So it is not the end of the world, I imagine this is also question of testing.

PR #2670, opened 2026-02-09
fix(config): disable open-in-view for Hibernate to fix Open-Session-in-View anti-pattern

balazs-szucs:
## 📝 Description

This PR disables Open Session In View (OSIV) to improve application performance and resource management. To support this change, the data access layer has been refactored to use explicit fetching strategies (Entity Graphs and Join Fetches), ensuring all required associations are loaded within the transactional boundary.

spring.jpa.open-in-view=true (default) keeps database connections open during entire HTTP request/response cycle:

- Causes N+1 query problems when lazy-loaded entities trigger queries during view rendering
- Increases risk of connection pool exhaustion and reduces throughput
- Violates separation of concerns by allowing SQL queries outside service layer

If you want to read more on this: https://martinelli.ch/the-hidden-performance-killer-understanding-open-session-in-view-in-spring-boot/ is great article that goes through the problem in really detailed way.

## 🏷️ Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Enhancement to existing feature
- [ ] Refactor (no behavior change)
- [ ] Breaking change (existing functionality affected)
- [ ] Documentation update

#### 1. Core Configuration

* Disabled OSIV: Set `spring.jpa.open-in-view: false` in `application.yaml`.

#### 2. Data Access Layer (Repositories)

* Optimized User Fetching: Introduced `findByIdWithDetails`, `findByIdWithSettings`, and `findByIdWithLibraries` in `UserRepository` using `@EntityGraph` to eagerly load permissions, settings, and library mappings in a single query.
* Optimized Book Fetching: Added `findByIdFull` and `findByIdWithMetadata` to `BookRepository` to handle complex metadata hierarchies (authors, tags, categories) efficiently.
* Batch Loading: Added `@BatchSize(size = 20)` to several `@OneToMany` and `@ManyToMany` collections in `BookEntity` and `BookLoreUserEntity` as a safety net for collection loading.

#### 3. Service & Security Refactoring

* Transactional Integrity: Added `@Transactional` annotations to various service methods (e.g., `ShelfService`, `UserService`, `LibraryService`) that previously relied on the OSIV filter to keep the Hibernate session open.
* Authentication Filters: Refactored security filters (JWT, Kobo, OIDC) to use the new "WithDetails" repository methods, preventing lazy loading errors during the authentication principal construction.
* Library Background Processing: Updated `LibraryService` to use `TransactionSynchronizationManager` to ensure background library scans only trigger *after* the initial transaction has successfully committed.

#### 4. API & Controller Cleanup

* Metadata Controller: Refactored `updateMetadata` logic into the `BookMetadataService` to keep controller logic thin and ensure transactionality.

### Impact

* Performance: Reduced the number of database round-trips ( problems) across the application.
* Stability: Eliminates the risk of `LazyInitializationException` in the view layer and prevents "hidden" database queries from occurring outside of service logic.
* Resource Management: Database connections are now released immediately after the service layer completes, rather than being held open until the request is fully rendered.

## 📸 Screenshots _(if applicable)_

---

## ✅ Pre-Submission Checklist

> **All boxes must be checked before requesting review.** Incomplete PRs will be closed without review, no exceptions.

- [X] Code follows project style guidelines and conventions
- [X] Branch is up to date with `develop` (merge conflicts resolved)
- [X] Automated tests added or updated to cover changes (backend **and** frontend)
- [X] All tests pass locally (`./gradlew test` for backend, `ng test` for frontend)
- [X] Changes manually verified in local dev environment (including related features)
- [ ] Flyway migration versioning is correct _(if schema was modified)_
- [ ] Documentation PR submitted to [booklore-docs](https://github.com/booklore-app/booklore-docs) _(if user-facing changes)_
- [ ] PR is reasonably scoped (PRs over 500+ changed lines will be closed, split into smaller PRs)

### 🤖 AI-Assisted Contributions

> **If any part of this PR was generated or assisted by AI tools (Copilot, Claude, ChatGPT, etc.), all items below are mandatory.** AI-generated code that hasn't been thoroughly reviewed is the leading cause of reverted PRs. You are responsible for every line you submit. "The AI wrote it" is not an excuse.

- [ ] Every line has been read, understood, and can be explained by the author
- [ ] PR is scoped to a single logical change, not a dump of everything the AI suggested
- [ ] Tests validate actual behavior, not just coverage (AI-generated tests often assert nothing meaningful)
- [ ] No dead code, placeholder comments, `TODO`s, or unused scaffolding

---

### DIFF
--- a/booklore-api/src/main/java/org/booklore/service/reader/AudioMetadataService.java
+++ b/booklore-api/src/main/java/org/booklore/service/reader/AudioMetadataService.java
@@ -1,7 +1,14 @@
 package org.booklore.service.reader;
 
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.booklore.model.dto.AudiobookMetadata;
 import org.booklore.model.dto.response.AudiobookChapter;
 import org.booklore.model.dto.response.AudiobookInfo;
 import org.booklore.model.dto.response.AudiobookTrack;
@@ -9,8 +16,6 @@ import org.booklore.model.entity.BookFileEntity;
 import org.booklore.model.entity.BookMetadataEntity;
 import org.booklore.repository.BookFileRepository;
 import org.booklore.service.metadata.extractor.AudiobookMetadataExtractor;
-import lombok.RequiredArgsConstructor;
-import lombok.extern.slf4j.Slf4j;
 import org.jaudiotagger.audio.AudioFile;
 import org.jaudiotagger.audio.AudioFileIO;
 import org.jaudiotagger.audio.AudioHeader;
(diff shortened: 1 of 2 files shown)


balazs-szucs:
Hi @acx10,

Another day, another large refactor. I'm afraid this one is quite a bit riskier than the previous migration (the PR with Java 25, Spring 4, etc.), so this will require quite a bit of testing. As far as reliability and performance gains go, this is quite significant.

I have a very strong suspicion that this will make orphaned DB entries a thing of the past as well. A rather large number of tickets are about that, and while running cleanup tasks is technically a workaround, it is very sub-optimal. I tested as much as I could, but please do not treat this as 100% working until it is proven so. I would like to squeeze this in before the release if possible, because this is only going to get worse and worse with more and more lazy-loaded entities in the codebase

acx10:
@balazs-szucs Yeah, I’ll take a look at it soon. Thanks for the context!

balazs-szucs:
Case in point here: https://github.com/booklore-app/booklore/issues/2508 :cry:

acx10:
I was looking into spring.jpa.open-in-view; setting it to false should help prevent lazy-loading issues from leaking into the controller or view layer. Were you able to test all the flows? We want to make sure there are no surprises for the user.

balazs-szucs:
> I was looking into spring.jpa.open-in-view; setting it to false should help prevent lazy-loading issues from leaking into the controller or view layer.

Yes. This is what this PR does. Though, it can be taken even a step further by tweaking DTO etc... But I wanted to keep this _more_ focused.

> Were you able to test all the flows? We want to make sure there are no surprises for the user.

As eladuded by: https://github.com/booklore-app/booklore/pull/2670#issuecomment-3871463297, I tried to almost everything I could. I am using this as my daily driver to see whether are there problem; however these changes as also suggested by whopping 82 files modified are fundamentally change the codebase, and it is rather hard to think through every and all implications or features that may be effected. As far as the everyday UX goes, there shouldnt be problems, but I cannot make any claims about the completeness of the testing simply because of how much these change things.

acx10:
Thanks, I’m always a bit cautious with PRs that touch so many files, since fully testing everything usually requires a complete manual regression, which can be tough. I’ll also try to go through it myself to verify things.

booklore-api/src/main/java/org/booklore/service/recommender/BookRecommendationService.java
```diff
@@ -9,18 +11,18 @@
...
 
-import java.util.*;
 import java.util.AbstractMap.SimpleEntry;
+import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
 @Slf4j
 @AllArgsConstructor
 @Service
```
acx10, line 24: This will break `getRecommendations()`. That method calls `bookRepository.save(book)` at line 42 to persist computed recommendations. With `readOnly = true` at the class level, Hibernate won't flush the write. Recommendations will be recomputed on every call but never stored.

Needs a method-level `@Transactional` override on `getRecommendations()`.

booklore-api/src/main/java/org/booklore/repository/BookRepository.java
```diff
@@ -24,6 +24,32 @@ public interface BookRepository extends JpaRepository<BookEntity, Long>, JpaSpec
...
+        LEFT JOIN FETCH m.categories
+        LEFT JOIN FETCH m.moods
+        LEFT JOIN FETCH m.tags
+        LEFT JOIN FETCH m.comicMetadata
+        WHERE b.id = :id AND (b.deleted IS NULL OR b.deleted = false)
+        """)
+    Optional<BookEntity> findByIdWithMetadata(@Param("id") Long id);
+
+    @Query("""
+        SELECT DISTINCT b FROM BookEntity b
```
acx10: 7 LEFT JOIN FETCHes in one query creates a Cartesian product. For a book with 3 authors, 5 categories, 2 moods, and 10 tags, this returns 300 duplicate rows that `DISTINCT` has to collapse. Gets worse with more metadata.

Consider splitting into two queries (one for metadata associations, one for bookFiles) or relying on `@BatchSize` instead of fetch joins for the collections.

booklore-api/src/main/java/org/booklore/config/security/aspect/BookAccessAspect.java
```diff
@@ -44,7 +46,7 @@ public void checkBookAccess(JoinPoint joinPoint) {
             throw ApiError.GENERIC_BAD_REQUEST.createException("Missing or invalid book ID in method parameters.");
```
acx10, line 46: This method only checks `book.getLibrary().getId()` and runs content restrictions. It never touches `bookFiles`. `findByIdWithBookFiles` eagerly loads metadata, comicMetadata, shelves, libraryPath, library, AND bookFiles. That's a lot of data for an access check.

A simple `findById` with just the `library` association fetched would be enough here.

booklore-api/src/main/java/org/booklore/service/book/BookReviewService.java
```diff
@@ -39,7 +39,7 @@ public class BookReviewService {
     private final AuthenticationService authenticationService;
 
     public List<BookReview> getByBookId(Long bookId) {
-        BookEntity bookEntity = bookRepository.findById(bookId).orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(bookId));
+        BookEntity bookEntity = bookRepository.findByIdWithBookFiles(bookId).orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(bookId));
```
acx10: Neither `getByBookId()` nor `refreshReviews()` access `bookFiles`. They only touch `metadata` and `reviews`. `findByIdWithBookFiles` loads a lot more than needed here.

booklore-api/src/main/java/org/booklore/service/hardcover/HardcoverSyncService.java
```diff
@@ -80,7 +80,7 @@ public void syncProgressToHardcover(Long bookId, Float progressPercent, Long use
                 }
 
                 // Fetch book fresh within the async context to avoid lazy loading issues
-                BookEntity book = bookRepository.findById(bookId).orElse(null);
+                BookEntity book = bookRepository.findByIdWithBookFiles(bookId).orElse(null);
```
acx10: This method only reads `metadata.getIsbn13()`, `metadata.getIsbn10()`, and `metadata.getHardcoverBookId()`. It never accesses `bookFiles`. Loading the full entity graph here is unnecessary overhead, especially since this runs on every progress sync.

booklore-api/src/main/java/org/booklore/service/library/LibraryService.java
```diff
@@ -145,26 +148,23 @@ public Library updateLibrary(CreateLibraryRequest request, Long libraryId) {
...
-                    libraryProcessingService.processLibrary(libraryId);
-                } catch (InvalidDataAccessApiUsageException e) {
-                    log.debug("InvalidDataAccessApiUsageException - Library id: {}", libraryId);
-                } finally {
-                    scanningLibraries.remove(libraryId);
-                }
-                log.info("Parsing task completed!");
-            });
+            if (TransactionSynchronizationManager.isSynchronizationActive()) {
+                TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
```
acx10: This block is duplicated in `createLibrary()` below. Worth extracting into a shared helper:
```java
private void scheduleBackgroundScanAfterCommit(long libraryId) {
    if (TransactionSynchronizationManager.isSynchronizationActive()) {
        TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
            @Override
            public void afterCommit() {
                startBackgroundScan(libraryId);
            }
        });
    } else {
        startBackgroundScan(libraryId);
    }
}
```

booklore-api/src/main/java/org/booklore/repository/BookRepository.java
```diff
@@ -24,6 +24,32 @@ public interface BookRepository extends JpaRepository<BookEntity, Long>, JpaSpec
     @Query("SELECT b FROM BookEntity b LEFT JOIN FETCH b.bookFiles bf WHERE b.id = :id AND (b.deleted IS NULL OR b.deleted = false)")
     Optional<BookEntity> findByIdWithBookFiles(@Param("id") Long id);
```
acx10, line 25: Minor: this query uses both `@EntityGraph(attributePaths = {..., "bookFiles"})` and `LEFT JOIN FETCH b.bookFiles bf`. Both achieve the same thing for `bookFiles`. The `@EntityGraph` already handles the eager fetch, so the `LEFT JOIN FETCH` in the JPQL is redundant.

booklore-api/src/main/java/org/booklore/service/user/UserService.java
```diff
@@ -79,19 +85,21 @@ public void deleteUser(Long id) {
...
-        BookLoreUserEntity user = userRepository.findById(id).orElseThrow(() -> ApiError.USER_NOT_FOUND.createException(id));
+        BookLoreUserEntity user = userRepository.findByIdWithDetails(id).orElseThrow(() -> ApiError.USER_NOT_FOUND.createException(id));
         return bookLoreUserTransformer.toDTO(user);
     }
 
     public BookLoreUser getMyself() {
         return authenticationService.getAuthenticatedUser();
     }
 
+    @Transactional
```
acx10, line 98: `changePassword()` only checks `permissions.isDemoUser()`. It doesn't need settings. `findByIdWithSettings` loads the full settings collection for nothing. Same applies to `changeUserPassword()` below.

balazs-szucs:
Hi, updated the PR. Retested whatever I could. Same philosophy as earlier; I'm afraid I couldn't test everything, main UX should be good.

PR #2678, opened 2026-02-09
feat(hardcover): get Hardcover metadata from a specific edition when ISBN is provided

pedronave:
## 📝 Description

Currently, when we search for metadata by ISBN, metadata providers like Goodreads provide the metadata for the edition with that ISBN. Hardcover, however, returns the default metadata of the book, leading to inaccurate metadata.

This change improves the Hardcover metadata search so that when an ISBN is provided, the metadata returned by the search matches the metadata associated with that ISBN on Hardcover.

## 🏷️ Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement to existing feature
- [ ] Refactor (no behavior change)
- [ ] Breaking change (existing functionality affected)
- [ ] Documentation update

## 🔧 Changes

- When an ISBN is provided, we get the specific edition(s) with that ISBN, and return the metadata from it
- Adds language and publisher metadata fetch
- Gets the actual ISBN10 and ISBN13 from that edition. If one is provided and the other not, calculates it (as was already the case)
- One thing to note is that this changes gives preference to correct metadata versus complete metadata. For example, if the edition does not have a pages count, it will not be returned with the metadata, even if the book has one (from the default edition).
  - Some metadata still comes from the book itself, like the description, as the edition does not have this data.

## 🧪 Testing

- Mapping of metadata from a book and its edition
- Mapping of multiple editions within the same book (should not happen as ISBN is unique, but could happen)
- Missing metadata does not break the mapping
- Null or empty results from the ISBN query do not break the mapping

## 📸 Screenshots _(if applicable)_

Before:
image
image

After:
image

---

## ✅ Pre-Submission Checklist

> **All boxes must be checked before requesting review.** Incomplete PRs will be closed without review, no exceptions.

- [x] Branch is up to date with `develop` (merge conflicts resolved)
- [x] Automated tests added or updated to cover changes (backend **and** frontend)
- [x] All tests pass locally (`./gradlew test` for backend, `ng test` for frontend)
- [x] Changes manually verified in local dev environment (including related features)
- [ ] Flyway migration versioning is correct _(if schema was modified)_
- [ ] Documentation PR submitted to [booklore-docs](https://github.com/booklore-app/booklore-docs) _(if user-facing changes)_
- [ ] PR is reasonably scoped (PRs over 500+ changed lines will be closed, split into smaller PRs)
  A bit more than 500, but mostly models, query and tests. The change itself is significantly smaller. If needed, I can try to break it down a bit

### 🤖 AI-Assisted Contributions

> **If any part of this PR was generated or assisted by AI tools (Copilot, Claude, ChatGPT, etc.), all items below are mandatory.** AI-generated code that hasn't been thoroughly reviewed is the leading cause of reverted PRs. You are responsible for every line you submit. "The AI wrote it" is not an excuse.

- [ ] Every line has been read, understood, and can be explained by the author
- [ ] PR is scoped to a single logical change, not a dump of everything the AI suggested
- [ ] Tests validate actual behavior, not just coverage (AI-generated tests often assert nothing meaningful)
- [ ] No dead code, placeholder comments, `TODO`s, or unused scaffolding

---


booklore-api/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
```diff
@@ -183,6 +204,102 @@ private BookMetadata mapDocumentToMetadata(GraphQLResponse.Document doc, FetchMe
...
+            metadata.setHardcoverBookId(bookId.toString());
+        }
+
+        metadata.setTitle(edition.getTitle());
+        metadata.setSubtitle(edition.getSubtitle());
+        metadata.setDescription(book.getDescription());
+
+        if (edition.getCachedContributors() != null) {
+            metadata.setAuthors(edition.getCachedContributors().stream()
+                    .map(GraphQLResponse.Contributor::getAuthor)
```
acx10, line 225: This will throw an NPE if any `Contributor` in the list has a null `author` field. The outer check confirms `cachedContributors` isn't null, but individual entries could still have `author == null` (incomplete data from the API, etc.). A `.filter(Objects::nonNull)` after each `.map()` would make this safe:

```java
metadata.setAuthors(edition.getCachedContributors().stream()
    .map(GraphQLResponse.Contributor::getAuthor)
    .filter(Objects::nonNull)
    .map(GraphQLResponse.Author::getName)
    .filter(Objects::nonNull)
    .collect(Collectors.toSet()));
```

booklore-api/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -35,6 +35,77 @@ public HardcoverBookSearchService(AppSettingService appSettingService) {
...
+
+        GraphQLResponse response = executeRequest(body, GraphQLResponse.class, apiToken);
+        if (response == null || response.getData() == null ||
+                response.getData().getBooks() == null ||
+                response.getData().getBooks().isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        List<GraphQLResponse.BookWithEditions> books = response.getData().getBooks();
+        return books != null ? books : Collections.emptyList();
```
acx10: This `books != null` check is redundant. Lines 100-101 already return early if `getBooks()` is null or empty. By the time you get here, `books` is guaranteed to be non-null and non-empty, so this can just be `return response.getData().getBooks();` (or `return books;` if you keep the local variable for clarity).

booklore-api/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -35,6 +35,77 @@ public HardcoverBookSearchService(AppSettingService appSettingService) {
...
+                          title
+                          subtitle
+                          cached_contributors
+                          pages
+                          release_date
+                          release_year
+                          image {
+                            url
+                          }
+                          cached_tags
```
acx10: Tiny thing: `cached_tags` is being fetched here on the edition, but `mapEditionToMetadata` only ever reads `book.getCachedTags()`, not `edition.getCachedTags()`. So the edition-level tags are coming back from the API but getting ignored. Either drop `cached_tags` from the edition part of the query to save a bit of bandwidth, or if editions can have their own tags that should take priority, use them in the mapper.

booklore-api/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
```diff
@@ -79,6 +79,27 @@ public List<BookMetadata> fetchMetadata(Book book, FetchMetadataRequest fetchMet
...
+        if (books == null || books.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        List<BookMetadata> results = new ArrayList<>();
+        for (GraphQLResponse.BookWithEditions book : books) {
+            if (book.getEditions() == null || book.getEditions().isEmpty()) {
+                continue;
+            }
+            for (GraphQLResponse.Edition edition : book.getEditions()) {
```
acx10, line 92: This logs at INFO for every edition processed. For ISBN search it's usually just one, so it's fine in practice. But if this method ever gets reused or if the API returns multiple editions, it could get chatty. Might be worth using `log.debug` here instead, just so it doesn't show up in normal logs.

booklore-api/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
pedronave: Good catch, fixed this

booklore-api/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
pedronave: Removed the local variable and unnecessary null check

booklore-api/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
pedronave: I removed the cached_tags from the edition in both the query and the model

booklore-api/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
pedronave: Switched this to a debug log

booklore-api/src/main/java/org/booklore/service/metadata/parser/hardcover/GraphQLResponse.java
```diff
@@ -189,4 +266,61 @@ public static class Series {
...
+    }
+
+    @Getter
+    @Setter
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public static class CachedTags {
+        @JsonProperty("Genre")
+        private List<HardcoverBookDetails.CachedTag> genre;
```
acx10: this ties `GraphQLResponse` to `HardcoverBookDetails.CachedTag` pretty tightly. if `CachedTag` ever changes shape over there it could silently break deserialization here. might be worth having a local `CachedTag` in `GraphQLResponse` instead, or pulling out a shared type both can use?

pedronave: At first, I created a local cached tag, but since this is a JSON object that is returned by the API (we can't query its properties specifically), I ended up pulling it out into a shared type. I think the data model around hardcover as a whole could use a revamp

PR #2706, opened 2026-02-12
fix: cap JVM heap at 384 MB to reduce memory usage

acx10:
## 📝 Description

Replace `MaxRAMPercentage=75.0` with an explicit `-Xmx384m` heap cap. Without a container memory limit, the JVM was claiming 75% of host RAM (e.g. ~12 GB on a 16 GB machine), resulting in ~800 MB+ resident memory for a self-hosted app that needs far less.

## 🏷️ Type of Change

- [x] Bug fix

## 🔧 Changes

- `Dockerfile`: replaced `-XX:MaxRAMPercentage=75.0` with `-Xmx384m`
- `Dockerfile.ci`: same change

## 🧪 Testing

- Verified the JVM respects the 384 MB cap via `docker stats` and actuator metrics
- Users can override via `JAVA_TOOL_OPTIONS` env var in their compose file if needed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ LABEL org.opencontainers.image.title="BookLore" \
       org.opencontainers.image.licenses="GPL-3.0" \
       org.opencontainers.image.base.name="docker.io/library/eclipse-temurin:25-jre-alpine"
 
-ENV JAVA_TOOL_OPTIONS="-XX:+UseG1GC -XX:MaxGCPauseMillis=200 -XX:+UseStringDeduplication -XX:+UseContainerSupport -Xmx384m -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:G1PeriodicGCInterval=60000"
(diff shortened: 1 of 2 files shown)


janpfischer:
Just my 2 cents but I am not sure about this change. I think it is dangerous to limit resources like this as hard values in the Image file. Wouldn't it be better to advise the users to limit the resources Docker assigns to the Container? This is possible via CLI arguments or within the Docker Compose file. 
For environments like Kubernetes this is not a problem, too because you can limit resources there too and it is best practice to do it in configuration not as hard values in the Image. 

https://docs.docker.com/engine/containers/resource_constraints/

balazs-szucs:
> Just my 2 cents but I am not sure about this change. I think it is dangerous to limit resources like this as hard values in the Image file. Wouldn't it be better to advise the users to limit the resources Docker assigns to the Container? This is possible via CLI arguments or within the Docker Compose file. For environments like Kubernetes this is not a problem, too because you can limit resources there too and it is best practice to do it in configuration not as hard values in the Image.
> 
> https://docs.docker.com/engine/containers/resource_constraints/

JVM heap limits vs. container memory limits are different things. Both can coexist, and setting a conservative JVM default doesn't preclude users from also setting container limits. In-fact it's quite common practice to do both. They are not "drop-in" replacement for each other. You may also override this option with your heap limit,  WHILE using a container mem limit.

janpfischer:
You are right that these are different things and they can coexist. I would be too afraid that this setting is too deep "buried" in the image to users to notice. Maybe a hint in the documentation and usage about Docker would be helpful here? With both notes on how to use limits in Docker and on howto overwrite this if needed.

PR #2712, opened 2026-02-12
feat(ui): add file name column to table view and include filename in search

acx10:
Added a File Name column to the table view and made the search bar match against filenames too. So now you can actually find books by their file name on disk. Closes #2710.


AucaCoyan:
thank you!

Issue #2366, opened 2026-01-20
[Feature] Global Bookmarks List


chiguirepower:
### What's Your Idea?

Add a new sidebar menu section called **Bookmarks**, similar to **Shelves**.

**Core Concept**
- Shelves organize books
- Bookmark Lists organize bookmarked pages across books

**Functionality**

1. Users can create multiple Bookmark Lists (custom groups).
2. Users can add existing bookmarks to one or more Bookmark Lists.
3. Selecting a Bookmark List opens a right-side panel/table showing:
- Book title
- Bookmark label/title (optional)
- Page / location
- Date added (optional)

**Behavior**

Clicking a bookmark row should open the reader and jump directly to the bookmarked page.

![Image](https://github.com/user-attachments/assets/5587696c-64dc-4f5f-8b9e-5194f383833d)
![Image](https://github.com/user-attachments/assets/b1ba4d0b-8796-4118-9221-4e8ac4d53a5f)
![Image](https://github.com/user-attachments/assets/fc0d9c00-2fb2-441e-a5d9-949c4daa279d)
![Image](https://github.com/user-attachments/assets/f7c0f7e4-7438-4a44-a2d7-52338840358b)
![Image](https://github.com/user-attachments/assets/6993fa09-ff48-4394-9694-6842524ad768)
![Image](https://github.com/user-attachments/assets/9f5ef23c-f61b-4625-9a3b-13b5421472fa)

### Why Would This Be Helpful?

Centralizing bookmarks by **topic or area of interest** allows users to study concepts instead of hunting through books.

**Use Case 1: Recipes & Cooking Techniques**

Many cookbooks contain valuable recipes and techniques spread across dozens of books. While reading, users often bookmark:
- Cake recipes
- Bread-making techniques
- Soups or specialty dishes

Without organization, revisiting a specific topic (e.g., bread techniques) requires searching through every related book individually.
With **Bookmark Lists** (e.g., “Bread Making Techniques”, “Cupcake Recipes”), users can:

- Study a single subject across multiple books
- Quickly access all relevant bookmarked pages in one place
- Treat bookmarks as a curated knowledge collection, not isolated notes

**Use Case 2: Self-Help Research & Writing**

Authors and researchers often bookmark meaningful passages while reading across many self-help books. These snippets may relate to:

- Mindfulness
- Emotional healing
- Personal growth themes

After finishing a topic or project, having subject-based Bookmark Lists allows them to:
- Revisit key ideas without reopening each book
[...]

PR #2736, opened 2026-02-13
feat: add annotation notebook with server-side pagination

acx10:
New Notebook page that pulls together all your highlights, notes, and bookmarks from every book into one place. Has server-side pagination, text search, type filters, a book dropdown filter, and you can export everything to markdown. Added composite indexes on the annotation tables to keep it fast.


chiguirepower:
Is this the same as #2366 ? Global Bookmarks?

PR #2808, opened 2026-02-19
Fix book browser header title overflow on mobile

acx10:
Fixes the book browser header where long library/shelf names would run into the toolbar controls on small screens. The title now properly truncates with an ellipsis by fixing the flex overflow chain (adding overflow: hidden and min-width: 0 to parent containers). Also tightened up the icon gaps on mobile to give more breathing room.

Closes #2807


rkram5424:
Holy smokes you guys are fast!

Issue #2824, opened 2026-02-21
[Bug] Attach Book Files attempts to move files when Auto-Move Files on Metadata Update is disabled


iamwillbar:
### What Went Wrong?

I have my library on a read-only network mount, when I tried to merge a format into another format it tried to move the underlying files, which failed. This is unexpected when Auto-Move Files on Metadata Update is disabled.

### How Can We Reproduce This?

1. Go to settings and ensure Auto-Move Files on Metadata Update is disabled.
2. Ensure you have two books with different file types in the library.
3. Open the first book, select Attach to Another Book, and select the second book.

### What Should Have Happened?

The files from the first book are moved, even though they shouldn't have been.

### Any Ideas on How to Fix This? (Optional)

Proposed fix:
https://github.com/booklore-app/booklore/compare/develop...iamwillbar:booklore:fix-attach-files?expand=1

### Your Setup

develop branch

### Before Submitting

- [x] I've searched existing issues and confirmed this bug hasn't been reported yet

balazs-szucs:
AFAIK; PDF viewer is using PDF.js instead Foliate, hence the problem. The solution is to migrate it to Foliate and it is not really related to multi-format.

PR #2833, opened 2026-02-21
Add moveFiles option to attach/merge and per-book delete in duplicate merger

acx10:
Was seeing issues where attaching book files would unconditionally move files on disk, even when the moveFilesToLibraryPattern setting was disabled. This broke for users with read-only mounts or who just wanted to reorganize book records without touching the filesystem.

Replaced the old deleteSourceBooks boolean with a moveFiles flag on the attach request. When moveFiles is false, files stay in place and only the DB records are reassigned (using JPQL UPDATE to avoid Hibernate orphan removal conflicts). The default for the checkbox in both the attach dialog and duplicate merger is pulled from the global moveFilesToLibraryPattern setting so it does the right thing out of the box. Also added per-book delete checkboxes in the duplicate merger so users can selectively remove individual duplicates from a group instead of being forced to merge everything. Includes 30 unit tests covering validation, both file-move paths, result enrichment, and edge cases.

### DIFF
--- a/booklore-ui/src/app/features/book/components/book-browser/book-browser.component.ts
+++ b/booklore-ui/src/app/features/book/components/book-browser/book-browser.component.ts
@@ -915,11 +915,10 @@ export class BookBrowserComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   attachFilesToBook(): void {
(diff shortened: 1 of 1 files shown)


acx10:
Fixes https://github.com/booklore-app/booklore/issues/2824

PR #2886, opened 2026-02-23
Fix metadata settings not saving due to null deserialization error

acx10:
Fixes #2884. Users upgrading from older versions had metadata_match_weights stored in the DB without newer fields (like seriesTotal, doubanRating, etc). Jackson was choking on null values for primitive int fields, which caused settings reads and saves to silently fail and fall back to defaults every time.

Added @JsonSetter(nulls = Nulls.SKIP) with @Builder.Default on all MetadataMatchWeights fields so missing/null values just keep sensible defaults instead of blowing up. Same pattern already used in MetadataRefreshOptions.

### DIFF
--- a/booklore-api/src/main/java/org/booklore/model/dto/settings/MetadataMatchWeights.java
+++ b/booklore-api/src/main/java/org/booklore/model/dto/settings/MetadataMatchWeights.java
@@ -1,6 +1,8 @@
 package org.booklore.model.dto.settings;
 
 
+import com.fasterxml.jackson.annotation.JsonSetter;
+import com.fasterxml.jackson.annotation.Nulls;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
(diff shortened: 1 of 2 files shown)


smorbickles:
I added this comment to my original bug report, but since that issue is closed, I'm not sure if it's monitored.
https://github.com/booklore-app/booklore/issues/2884#issuecomment-3953304171

> I'm getting the same errors after updating to v2.0.1.
> 
> ```
> booklore          | 2026-02-24T08:27:18.862-08:00 ERROR 1 --- [booklore-api] [cat-handler-355] o.b.s.a.SettingPersistenceHelper         : Failed to parse JSON for setting key 'metadata_match_weights'. Using default value. Error: Cannot map `null` into type `int` (set DeserializationConfig.DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES to 'false' to allow)
> booklore          |  at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN] (through reference chain: org.booklore.model.dto.settings.MetadataMatchWeights["seriesTotal"])
> booklore          | 2026-02-24T08:27:25.421-08:00 ERROR 1 --- [booklore-api] [cat-handler-357] o.b.s.a.SettingPersistenceHelper         : Failed to parse JSON for setting key 'metadata_match_weights'. Using default value. Error: Cannot map `null` into type `int` (set DeserializationConfig.DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES to 'false' to allow)
> booklore          |  at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN] (through reference chain: org.booklore.model.dto.settings.MetadataMatchWeights["seriesTotal"])
> booklore          | 2026-02-24T08:27:26.301-08:00 ERROR 1 --- [booklore-api] [cat-handler-359] o.b.s.a.SettingPersistenceHelper         : Failed to parse JSON for setting key 'metadata_match_weights'. Using default value. Error: Cannot map `null` into type `int` (set DeserializationConfig.DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES to 'false' to allow)
> booklore          |  at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN] (through reference chain: org.booklore.model.dto.settings.MetadataMatchWeights["seriesTotal"])
> ```
> 
> I believe [#2886](https://github.com/booklore-app/booklore/pull/2886) helps new values, but doesn't clean up the old broken ones. My database still has the old invalid metadata setting from 1.18.5 (seriesTotal is null). I think for migration from 1.18.5 to 2.0.0+, Booklore should detect the invalid null value and overwrite it with a valid default value.

PR #2950, opened 2026-02-25
Fix Docker build for Alpine 3.23 unrar removal

acx10:
Alpine 3.23 dropped the `unrar` package entirely, so the Docker build was failing. This vendors pre-built unrar 7.2.4 binaries (amd64 + arm64, compiled from RARLAB source against Alpine 3.23 musl) directly into the repo and COPYs the right one via TARGETARCH. No more network fetches during builds.

Also adds integration tests that feed a real RAR5 archive through the full fallback chain (junrar fails, unrar CLI picks up) across CbxReaderService, CbxMetadataExtractor, CbxMetadataWriter, and CbxConversionService.


NamLe029:
I think we could use unrar binary from this linuxserver image:
https://hub.docker.com/r/linuxserver/unrar

PR #2976, opened 2026-02-25
Fix audiobook chapter extraction for M4B/M4A formats

acx10:
Chapter extraction was broken for most audiobook formats, especially M4B/M4A. The OS lock screen would just show "Full Audiobook" instead of actual chapter names because ffprobe wasn't available in the Docker image, and the read-time fallback was hardcoding a dummy chapter instead of trying real extraction.

This adds on-demand ffprobe binary download (same pattern as kepubify), wires it into the metadata extractor so M4B chapter atoms actually get parsed, and adds lazy backfill so existing audiobooks get their chapters populated the next time they're opened. Rescan also updates audiobook technical metadata now instead of silently discarding it.

The frontend change makes the media session show the book title when there's only one chapter, so the lock screen isn't stuck on "Full Audiobook".

### DIFF
--- a/booklore-api/.gitignore
+++ b/booklore-api/.gitignore
@@ -35,3 +35,6 @@ out/
 
 ### VS Code ###
 .vscode/
+
+### Local scripts ###
+local/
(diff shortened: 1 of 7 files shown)


acx10:
closes #2915

Issue #2977, opened 2026-02-25
[Bug] Page (still) turning in Scroll mode (Mobile; Built-in reader)


Bro0528:
### What Went Wrong?

In Scroll mode, the page can be turned by swiping horizontally right or left.
This leads to random page flips when you're not expecting them and prevents smooth scrolling.
This makes using the built-in reader on the mobile version annoying.

In Scroll mode, there should be only vertical swiping for reading.

### How Can We Reproduce This?

1. Open a book in the mobile web reader.

2. Go to Settings → Scroll mode.

3. Swipe right or left.

### What Should Have Happened?

The page should not be turned.

### Screenshots or Error Messages (Optional)

https://github.com/user-attachments/assets/cda110bb-1c2d-4404-b95f-d4b9df6da516

### Any Ideas on How to Fix This? (Optional)

Just lock page swiping in scroll mode :)

https://github.com/booklore-app/booklore/issues/2154
I've already encountered this bug before; now to reproduce it, you need to swipe instead of touch.

### Your Setup

v.2.0.1
Docker

### Before Submitting

- [x] I've searched existing issues and confirmed this bug hasn't been reported yet

PR #3026, opened 2026-02-26
Disable horizontal swipe navigation in CBX infinite/long strip scroll modes (#2977)

acx10:
In the CBX reader, horizontal swipes were triggering nextPage()/previousPage() even in infinite and long strip scroll modes, which don't have discrete pages to navigate between. Now swipe gestures are skipped entirely in those modes.


Bro0528:
@acx10 Hi!

I didn't quite understand your response on this (2977) issue. It seems you gave an answer and closed the bug with a "won't fix" resolution, but I see there's a mention of it in the changelog for version 2.0.3.

If the fix was indeed supposed to be included, I don't have it and the page swipes are still not blocked.

PR #3127, opened 2026-03-01
Turn off auto book search by default (#3117)

### LINKED ISSUE #3117: [Enhancement] Turn off "Auto Book Search" by default
### What's Your Idea?

Currently, the "Auto Book Search" feature fetches metadata automatically whenever a book is opened in the Metadata Center. This setting is turned ON by default, so users start to hit API rate limits while using the app even before they discover this setting.

Proposed changes:
1. Turn OFF by default.
2. If turned ON, only trigger the metadata fetch when a user explicitly goes on the "Search Metadata" tab.
3. If turned ON, add an indicator to remind users of their current setting: e.g., Auto book search is ON. (This is a nice-to-have, refer to #3087)

### Why Would This Be Helpful?

This reduces unnecessary API calls that a user didn't intend to make. Ideally, users should have control over when  API calls are made to metadata providers. By turning this setting on by default without users' knowledge, it may cause friction down the line in other areas.

This may help reduce the Google Books API rate limit and Amazon service unavailable (503) errors that users get.

### Anything Else? (Optional)

_No response_

### Want to Help Out?

I'm happy to help test and give feedback

### Have You Considered Any Alternatives? (Optional)

Consider deprecating this feature (depending on its actual usage) in order to reduce unnecessary API calls and maintenance. Not too sure where else this feature is used.

### Before Submitting

- [x] I've searched existing issues and confirmed this feature hasn't been requested yet

acx10:
Flips the auto book search default to off so users don't hit API rate limits before they even know the setting exists. Also gates the auto-search to only fire when the Search Metadata tab is active, not on every book open.

Fixes #3117

### DIFF
--- a/booklore-api/src/main/java/org/booklore/service/appsettings/AppSettingService.java
+++ b/booklore-api/src/main/java/org/booklore/service/appsettings/AppSettingService.java
@@ -149,7 +149,7 @@ public class AppSettingService {
                 true
             )
         );
-        builder.autoBookSearch(Boolean.parseBoolean(settingPersistenceHelper.getOrCreateSetting(AppSettingKey.AUTO_BOOK_SEARCH, "true")));
+        builder.autoBookSearch(Boolean.parseBoolean(settingPersistenceHelper.getOrCreateSetting(AppSettingKey.AUTO_BOOK_SEARCH, "false")));
         builder.uploadPattern(settingPersistenceHelper.getOrCreateSetting(AppSettingKey.UPLOAD_FILE_PATTERN, "{authors}/<{series}/><{seriesIndex}. >/{title}/{title}< - {authors}>< ({year})>"));
         builder.similarBookRecommendation(Boolean.parseBoolean(settingPersistenceHelper.getOrCreateSetting(AppSettingKey.SIMILAR_BOOK_RECOMMENDATION, "true")));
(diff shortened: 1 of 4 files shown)


balazs-szucs:
So, this makes the metadata search more painful, with essentially zero benefits. Obviously, not my monkeys, not my circus, but take the covers here, for instance: #1963. 

If we won't get tickets about this shortly after release, I would be really surprised... but ok.

amara-vo:
> So, this makes the metadata search more painful, with essentially zero benefits. 

Could you clarify what you mean? As in, it'd require more steps for a user to search metadata? Or that this change affects other areas of searching metadata on the backend?

From my understanding (only looking at the UI), even before this commit was made, the Auto Book Search only affects the "Search Metadata" tab, not the whole Metadata Center. (Would love if @acx10 could clarify where else Auto Book Search is used.)

For #1963, from my manual testing, those two behaviors are independent of each other. I have Auto Book Search turned off, but the "Search for cover" button still triggers a search and populates covers upon clicking on it. This is in line with my expectations from a user perspective.

@balazs-szucs

balazs-szucs:
No need to spin a whole argument out of this, frankly. I have zero desire for that, I just wanted to get this out there. I think my points are clear.

PR #3165, opened 2026-03-04
Overhaul OIDC authentication

acx10:
Rewrite the OIDC flow to use a backend token exchange (BFF-lite) pattern. Previously, OIDC tokens were handled in the browser with a monolithic dual JWT filter. Now the backend does the full code exchange, token validation, and session management, so tokens never touch the frontend.

Backend:
- New `oidc` package with dedicated services for auth, token exchange, validation, discovery, state management, and backchannel logout
- PKCE, nonce, and state validation on every exchange. At-hash verification, redirect URI allowlist, JWKS caching with TTL
- Per-user ReentrantLock to prevent race conditions during concurrent user provisioning
- OIDC session tracking with a cleanup cron (revoked after 7 days, orphaned after 30)
- Group claim mapping with sync modes (disabled, on-login replace, on-login additive)
- Test connection endpoint with provider diagnostics
- Local account linking option for OIDC users claiming existing accounts
- Removed `DualJwtAuthenticationFilter`, `DynamicOidcJwtProcessor`, `MobileOidcController`

Frontend:
- New `OidcService` handles PKCE generation, state storage, and code exchange via the backend
- OIDC-only mode with auto-redirect and circuit breaker
- Expanded auth settings UI: provider config, group mapping table, test connection, session controls
- Login page adapts between combined (password + OIDC) and OIDC-only layouts

Migrations: V127 (OIDC schema), V128 (group mapping table)


acx10:
OIDC setup guide is up at https://booklore.org/docs/authentication/oidc-settings

jadehawk:
@acx10 
Running test container:
```
ghcr.io/booklore-app/booklore:pr-3165-9de62078
```

So far is working perfectly with Authentik + Plex Login setup.
Tested logging in and out using Edge browser as well as on my phone (android, galaxy s24). Also tested the auto creation of user from both Edge and mobile browsers.

So far is perfect. it will be nice if other users can report back using other providers (Authelia, Pocket ID). 

I deployed a fresh DB and Fresh container running v2.0.5, before using this image.
On first run I was caught on a LOOP. and had to delete cookies and clear cache on browser. It was the only hiccup. 

Then Had to restart container once I confirmed OIDC setting where the same, and after that no problems..

Thank you for overhauling this.

acx10:
> @acx10 Running test container:
> 
> ```
> ghcr.io/booklore-app/booklore:pr-3165-9de62078
> ```
> 
> So far is working perfectly with Authentik + Plex Login setup. Tested logging in and out using Edge browser as well as on my phone (android, galaxy s24). Also tested the auto creation of user from both Edge and mobile browsers.
> 
> So far is perfect. it will be nice if other users can report back using other providers (Authelia, Pocket ID).
> 
> I deployed a fresh DB and Fresh container running v2.0.5, before using this image. On first run I was caught on a LOOP. and had to delete cookies and clear cache on browser. It was the only hiccup.
> 
> Then Had to restart container once I confirmed OIDC setting where the same, and after that no problems..
> 
> Thank you for overhauling this.

@jadehawk Thanks for testing! Others in the thread have also confirmed it working with Zitadel, Authentik, and Authelia, and I've verified it works with Pocket ID as well. I think we're good to merge into develop, let it sit for a couple of days, and then cut a release.

PR #3282, opened 2026-03-11
Add BOOK_PER_FILE and BOOK_PER_FOLDER library organization modes

acx10:
Replaces the heuristic-based Auto-detect scanning with two explicit organization modes: Book per File (one file = one book) and Book per Folder (files in the same directory grouped together). Auto-detect was struggling with multi-file audiobooks and multi-format books, often missing files or splitting audiobooks into separate entries. The new modes are predictable, well-tested, and now the default for new libraries. Auto-detect still works for existing libraries but the dropdown is locked since switching away from it can't regroup existing books without a full wipe.

Also swaps the old event-based file monitoring for a WatchService implementation with move detection and a pending deletion pool, filters out temp/partial files and .ignore directories during scans, and adds a bunch of tests for the new grouping logic and file event processing.


altotin:
edit: moved to https://github.com/booklore-app/booklore/discussions/3299

PR #3312, opened 2026-03-13
Fix NPE when deleting authors (#3296)

acx10:
Deleting an author was causing a NullPointerException because the DB cascade delete left gaps in the sort_order column of the author mapping table. Hibernate interprets those gaps as null elements in the ordered list, so any subsequent book fetch that streams over authors would blow up. Now we remove the author from each book's author list before deleting, letting Hibernate handle the reindexing properly.

Fixes #3296


rcpeken:
Great catch! I reproduced this locally and actually opened a PR (#3311) to fix the NPE at the presentation layer by adding a defensive null check in the mapper. But fixing the root cause with the DB cascade and sort_order at the service layer like you did is definitely the better approach. I'll close my PR.

